PR #48, opened 2025-11-05
Add Serial Number to Avahi Service for Board Deduplication

mirkoCrobu:
### Motivation

A board connected via both USB and Network currently shows up as two separate devices in the board list. This makes it difficult for a client application to know it is the same physical board.

We need a way to "deduplicate" these entries by providing a common identifier, allowing the client to show just one entry for that board.

### Change description

A simple approach is to add the board's serial number to its Avahi service file. This change will be made at runtime when the EnableNetworkMode function is executed.

Specifically, we will add a new <txt-record> containing the serial number (e.g., <txt-record>serial_number=YOUR_SERIAL</txt-record>) to the board's .service file.

### Reviewer checklist

### DIFF
--- a/debian/arduino-app-cli/etc/systemd/system/arduino-avahi-serial.service
+++ b/debian/arduino-app-cli/etc/systemd/system/arduino-avahi-serial.service
@@ -6,7 +6,7 @@ ConditionPathExists=!/var/lib/arduino-app-cli/avahi_serial_configured.flag
 [Service]
 Type=oneshot
 RemainAfterExit=true
-ExecStart=/usr/local/bin/arduino-avahi-serial.sh
+ExecStart=/var/lib/arduino-app-cli/arduino-avahi-serial.sh
 ExecStartPost=/bin/mkdir -p /var/lib/arduino-app-cli
 ExecStartPost=/bin/touch /var/lib/arduino-app-cli/avahi_serial_configured.flag
 
--- a/debian/arduino-app-cli/var/lib/arduino-app-cli/arduino-avahi-serial.sh
+++ b/debian/arduino-app-cli/var/lib/arduino-app-cli/arduino-avahi-serial.sh
@@ -3,7 +3,6 @@
 
 
 TARGET_FILE="/etc/avahi/services/arduino.service"
-MARKER_LINE="</service>"
 SERIAL_NUMBER_PATH="/sys/devices/soc0/serial_number"
 
 echo "Configuring Avahi with serial number for network discovery..."
@@ -27,7 +26,7 @@ if [ -z "$SERIAL_NUMBER" ]; then
 fi
 
 if grep -q "serial_number=" "$TARGET_FILE"; then
-    echo "Serial number ($SERIAL_NUMBER) already configured. Skipping."
+    echo "Serial number ($SERIAL_NUMBER) already configured."
     exit 0
 fi
 


lucarin91 (changes requested):
I didn't figure out that we needed to add another sudoers file for this feature, which I would avoid to do it. So I think we should change the approach here, and we could instead do that in a systemd unit that is triggered only on the first boot, e.g., https://github.com/arduino/arduino-app-cli/blob/main/debian/arduino-app-cli/etc/systemd/system/arduino-burn-bootloader.service. 
The Serial number of the board should not change, so we can do it just the first time you flash the board, and in there, you have all the permission to patch the Avahi service.

mirkoCrobu:
### How to test it

1. Create the .deb and push it into the board
**On the board**
2. Be sure that  `/etc/avahi/services/arduino.service` doesn't contain the serial number
3. Reboot the board 
4. Check `/etc/avahi/services/arduino.service`: now you should see the serial number
5. Check that the .flag file exists:` ls -l /var/lib/arduino/avahi_serial_configured.flag`
6. Check the service status (should be active (exited)): `systemctl status arduino-avahi-serial.service`
7. Enable network mode: `arduino-app-cli system network-mode enable`
**From your PC** 
8. Check the board list configs(you should  the network information with serial number: 
): `arduino-cli board list --json`

debian/arduino-app-cli/etc/systemd/system/arduino-avahi-serial.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Configure Avahi with board serial number
+After=avahi-daemon.service
```
lucarin91: I would do it before not after

debian/arduino-app-cli/usr/sbin/arduino-avahi-serial.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+#
+# Configure Avahi with the serial number.
+# This operation is non-blocking: if it fails,
+# the script will exit with success in order to 
+# not to interrupt the post-install process.
```
lucarin91: I think this script should and could fail. If there is an issue, I would prefer that it start the next time instead of never

debian/arduino-app-cli/usr/sbin/arduino-avahi-serial.sh
```diff
@@ -0,0 +1,46 @@
...
+
+TARGET_FILE="/etc/avahi/services/arduino.service"
+MARKER_LINE="</service>"
+SERIAL_NUMBER_PATH="/sys/devices/soc0/serial_number"
+
+echo "Configuring Avahi with serial number for network discovery..."
+
+if [ ! -f "$SERIAL_NUMBER_PATH" ]; then
+    echo "Warning: Serial number path not found at $SERIAL_NUMBER_PATH. Skipping." >&2
+    exit 0
```
lucarin91:
```suggestion
    echo "Warning: Serial number path not found at $SERIAL_NUMBER_PATH" >&2
    exit 1
```

debian/arduino-app-cli/usr/sbin/arduino-avahi-serial.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo "Warning: Serial number file is empty. Skipping." >&2
+    exit 0 
+fi
+
+if grep -q "serial_number=${SERIAL_NUMBER}" "$TARGET_FILE"; then
+    echo "Serial number ($SERIAL_NUMBER) already configured. Skipping."
+    exit 0
+fi
+
+SERIAL_NUMBER_ESCAPED=$(echo "$SERIAL_NUMBER" | sed -e 's/\\/\\\\/g' -e 's/\//\\\//g' -e 's/\&/\\\&/g')
```
lucarin91: The serial number on the unoq is just a number I don't think you need to escape anything

debian/arduino-app-cli/DEBIAN/postinst
```diff
@@ -4,3 +4,4 @@ chown -R arduino:arduino /home/arduino/.local/share/arduino-app-cli
 
 systemctl enable arduino-app-cli
 systemctl enable arduino-burn-bootloader
+systemctl enable arduino-avahi-serial.service
```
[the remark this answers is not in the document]
mirkoCrobu: I test it and I get: 
`Error: Target file /etc/avahi/services/arduino not found or not writable.`

debian/arduino-app-cli/usr/sbin/arduino-avahi-serial.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+SERIAL_NUMBER=$(cat "$SERIAL_NUMBER_PATH")
+
+if [ -z "$SERIAL_NUMBER" ]; then
+    echo "Error: Serial number file is empty." >&2
+    exit 1 
+fi
+
+if grep -q "serial_number=${SERIAL_NUMBER}" "$TARGET_FILE"; then
```
lucarin91: I would do the check only on the `serial_number` properties to be sure we don't include multiple times the property `serial_number`

lucarin91 (commented):
LGTM, I would improve a little bit the schell script

debian/arduino-app-cli/etc/systemd/system/arduino-avahi-serial.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Configure Avahi with board serial number
+Before=avahi-daemon.service
+ConditionPathExists=!/var/lib/arduino/avahi_serial_configured.flag
```
dido18: Is there any reason why not to use the same folder location used by  `arduino-burn-bootloader.service` ?

It stores the flag in this path:
```
 /home/arduino/.local/share/arduino-app-cli/bootloader_burned.flag
```

dido18 (changes requested):

debian/arduino-app-cli/etc/systemd/system/arduino-avahi-serial.service
lucarin91: `arduino-burn-bootloader.service` runs as `arduino` user. This one is `root`. We don't want files with root permissions in the home directory of the `arduino` user.

mirkoCrobu:
@lucarin91 I had to change a commit author already includedd in main to pass license/cla. 
Let me check if I can fix it

debian/arduino-app-cli/usr/local/bin/arduino-avahi-serial.sh
```diff
@@ -0,0 +1,38 @@
...
+
+SERIAL_NUMBER=$(cat "$SERIAL_NUMBER_PATH")
+
+if [ -z "$SERIAL_NUMBER" ]; then
+    echo "Error: Serial number file is empty." >&2
+    exit 1 
+fi
+
+if grep -q "serial_number=" "$TARGET_FILE"; then
+    echo "Serial number ($SERIAL_NUMBER) already configured. Skipping."
```
lucarin91:
```suggestion
    echo "Serial number ($SERIAL_NUMBER) already configured."
```

lucarin91 (commented):
I think it is safe to always return errors

PR #50, opened 2025-11-06
fix: install a library missing from local library-index

### LINKED ISSUE #45: The installation of a library fails because thelibrary- index is not update
When a new library (listed in the library index) is added to the sketch.yaml file, the Arduino CLI fails to install it.

This is because the library index is not updated when a new library is added.

See https://forum.arduino.cc/t/good-morning-everyone-app-lab-0-2-0-doesnt-load-the-modulino-h-library/1411464

dido18:
### Motivation
Fix #45.

If a library is installed on the board, but it is not present in the local `library_index`, the installation fails because it is not found.

This can happen when a library is added to the public library registry, but the local index on the board is not updated.

The manual solution is to launch the command `arduino-cli lib update-index` on the board.

### Change description 
- When a library is added, if the local library index has expired, the local library index is updated by running `arduino-cli lib update-index`
-  The local library-index expires in `10 minutes` 

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/sketch_libs.go
+++ b/internal/orchestrator/sketch_libs.go
@@ -53,7 +53,6 @@ func AddSketchLibrary(ctx context.Context, app app.ArduinoApp, libRef LibraryRel
 	// update the local library index after a certain time, to avoid if a library is added to the sketch but the local library index is not update, the compile can fail (because the lib is not found)
 	req := &rpc.UpdateLibrariesIndexRequest{Instance: inst, UpdateIfOlderThanSecs: int64(indexUpdateInterval.Seconds())}
 	if err := srv.UpdateLibrariesIndex(req, stream); err != nil {
-		// ignore the error because a missing connection should not stop the user from adding a library.
 		slog.Warn("error updating library index, skipping", slog.String("error", err.Error()))
 	}
 


cmaglie (approved):

internal/orchestrator/sketch_libs.go
```diff
@@ -43,6 +48,16 @@ func AddSketchLibrary(ctx context.Context, app app.ArduinoApp, libRef LibraryRel
 		return nil, err
 	}
 
+	// update the local library_index if it is older than a certain threshold, to ensure the library is found when added by the arduino-cli
+	stream, res := commands.UpdateLibrariesIndexStreamResponseToCallbackFunction(ctx, func(curr *rpc.DownloadProgress) {
+		slog.Debug("downloading library index", "progress", curr.Message)
+	})
+	req := &rpc.UpdateLibrariesIndexRequest{Instance: inst, UpdateIfOlderThanSecs: int64(indexUpdateInterval.Seconds())}
+	if err := srv.UpdateLibrariesIndex(req, stream); err != nil {
+		return []LibraryReleaseID{}, fmt.Errorf("error updating library index: %v", err)
```
lucarin91: What happens if the board is disconnected? I think we should allow adding a library also without an internet connection

cmaglie: You will need the connection anyway to download the library.

PR #59, opened 2025-11-10
feat: implement app cache deletion

### LINKED ISSUE #52: add cache cleanup command
Arduino apps (and examples) have a .cache folder where both python/docker and the sketch add their compilation running temporary files. We should have a command to delete this data and force recreation. This could be useful boot for spare space and for solving some issue see https://github.com/arduino/arduino-app-cli/issues/51

alessio-perugini:
Introduce a `cache clean <app-id>` command to clean app cache. This implementation also tries to stop the related app if running. In case we do not care about that we can simply rm -rf `.cache/*`.

Super simple implementation. The stopping of an app could be opionionated, eager to see what you think. In case it's too much we can remove that logic and/or trigger it with a specific flag.

Closes #52

### DIFF
--- a/internal/orchestrator/cache.go
+++ b/internal/orchestrator/cache.go
@@ -32,7 +32,9 @@ func CleanAppCache(
 			return ErrCleanCacheRunningApp
 		}
 		// We try to remove docker related resources at best effort
-		_ = StopAndDestroyApp(ctx, app)
+		for range StopAndDestroyApp(ctx, app) {
+			// just consume the iterator
+		}
 	}
 
 	return app.ProvisioningStateDir().RemoveAll()


lucarin91:
I also not sure if the command should be in the `app` sub-command, for instance
```
arduino-app-cli app clean-cache <app-id> [--force]
```

alessio-perugini:
While implementing the CLI subcommands I took inspiration from the arduino-cli. My idea was to have similar mapping to what our user based is used to.

However I also feel like that the `cache`/`clean-cache` could be implemented as a subcommand of the `app`. Probably would be easier to understand it's scope, can be a bit surprising to have a generic `cache` that, for now, only operates on app related stuff.

> cache clean <app-id> --force stops the app and then deletes the cache

@lucarin91 yeah I like that

mirkoCrobu:
@alessio-perugini  Thinking about it again, `arduino-app-cli app clean-cache <app-id> [--force]` I agree with @lucarin91  
The clean-cache is at the app level. `System `is too general.

alessio-perugini:
@mirkoCrobu @lucarin91 137fcc3

internal/orchestrator/cache.go
```diff
@@ -0,0 +1,40 @@
...
+		return err
+	}
+	if runningApp.FullPath.EqualsTo(app.FullPath) {
+		return ErrCleanCacheRunningApp
+	}
+
+	if req.ForceClean && app.AppComposeFilePath().Exist() {
+		// We try to remove docker related resources at best effort
+		_ = StopAndDestroyApp(ctx, app)
+	}
```
lucarin91: You weren't handling the force flag correctly. I think this should work

```suggestion
    if runningApp.FullPath.EqualsTo(app.FullPath) {
       if req.ForceClean {
		// We try to remove docker related resources at best effort
		_ = StopAndDestroyApp(ctx, app)
	   }else{
		return ErrCleanCacheRunningApp
		}
	}
```

lucarin91 (changes requested):

internal/orchestrator/cache.go
```diff
@@ -0,0 +1,39 @@
...
+	ctx context.Context,
+	docker command.Cli,
+	app app.ArduinoApp,
+	req CleanAppCacheRequest,
+) error {
+	runningApp, err := getRunningApp(ctx, docker.Client())
+	if err != nil {
+		return err
+	}
+	if runningApp.FullPath.EqualsTo(app.FullPath) {
```
lucarin91: Sorry, I forgot to add a nil check here 😮‍💨 

```suggestion
	if runningApp != nil && runningApp.FullPath.EqualsTo(app.FullPath) {
```

internal/orchestrator/cache.go
```diff
@@ -0,0 +1,39 @@
...
+		if !req.ForceClean {
+			return ErrCleanCacheRunningApp
+		}
+		// We try to remove docker related resources at best effort
+		_ = StopAndDestroyApp(ctx, app)
```
lucarin91: This is an iterator, so you should consume it, otherwise it will not execute anything

```suggestion
		for range StopAndDestroyApp(ctx, app) {
			// just consume the iterator
		}
```

internal/orchestrator/cache.go
```diff
@@ -0,0 +1,41 @@
...
+	if runningApp != nil && runningApp.FullPath.EqualsTo(app.FullPath) {
+		if !req.ForceClean {
+			return ErrCleanCacheRunningApp
+		}
+		// We try to remove docker related resources at best effort
+		for range StopAndDestroyApp(ctx, app) {
+			// just consume the iterator
+		}
+	}
+
```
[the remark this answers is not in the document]
lucarin91: The app details is way more complicated, I think getRunningApp is enough in this case

PR #62, opened 2025-11-11
fix(update): return better errors

lucarin91:
### Motivation

We should improve the errors returned during the update operation, allowing the caller to be able to understand what happened and if to retry 

### Change description

1. Add a `code` field in the error structure, both in normal HTTP calls and in SSE ones, with a set of easy-to-match strings
2. Return the full underline error in the SEE error event and in the command line
3. code refactoring (hopping it is more redable)

The error codes for now are:
- `NO_INTERNET_CONNECTION`
- `OPERATION_IN_PROGRESS`
- `UNKNOWN_ERROR`

Both `NO_INTERNET_CONNECTION` and `OPERATION_IN_PROGRESS` could be considered retryable errors.

### Reviewer checklist

### DIFF
--- a/internal/api/handlers/update.go
+++ b/internal/api/handlers/update.go
@@ -43,7 +43,7 @@ func HandleCheckUpgradable(updater *update.Manager) http.HandlerFunc {
 		pkgs, err := updater.ListUpgradablePackages(r.Context(), filterFunc)
 		if err != nil {
 			code := update.GetUpdateErrorCode(err)
-			if code == update.OperationInProgress {
+			if code == update.OperationInProgressCode {
 				render.EncodeResponse(w, http.StatusConflict, models.ErrorResponse{
 					Code:    string(code),
 					Details: err.Error(),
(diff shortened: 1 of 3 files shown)


internal/update/apt/service.go
```diff
@@ -84,79 +83,55 @@ func (s *Service) UpgradePackages(ctx context.Context, names []string) (<-chan u
...
 			}
-			eventsCh <- update.Event{Type: update.UpgradeLineEvent, Data: line}
+			eventsCh <- update.NewDataEvent(update.UpgradeLineEvent, line)
 		}
-		// TEMPORARY PATCH: stopping and destroying docker containers and images since IDE does not implement it yet.
-		// TODO: Remove this workaround once IDE implements it.
-		// Tracking issue: https://github.com/arduino/arduino-app-cli/issues/623
-		eventsCh <- update.Event{Type: update.UpgradeLineEvent, Data: "Stop and destroy docker containers and images ..."}
+
+		eventsCh <- update.NewDataEvent(update.UpgradeLineEvent, "Apt upgrade completed successfully.")
```
dido18:
```suggestion
		eventsCh <- update.NewDataEvent(update.UpgradeLineEvent, "Stop and destroy docker containers and images ....")
```

martacarbone (approved):

PR #63, opened 2025-11-11
Expose serial number for App-Lab in network mode

mirkoCrobu:
### Motivation

A board connected via both USB and Network currently shows up as two separate devices in the board list. This makes it difficult for a client application to know it is the same physical board.

We need a way to "deduplicate" these entries by providing a common identifier, allowing the client to show just one entry for that board.

### Reviewer checklist

### DIFF
--- a/pkg/board/board.go
+++ b/pkg/board/board.go
@@ -182,9 +182,10 @@ func FromFQBN(ctx context.Context, fqbn string) ([]Board, error) {
 		if len(port.GetMatchingBoards()) > 0 {
 			boardName = port.GetMatchingBoards()[0].GetName()
 		}
-		serial := strings.ToLower(port.GetPort().GetHardwareId()) // in windows this is uppercase.
+
 		switch port.GetPort().GetProtocol() {
 		case SerialProtocol:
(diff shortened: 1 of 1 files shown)


pkg/board/board.go
```diff
@@ -215,6 +213,7 @@ func FromFQBN(ctx context.Context, fqbn string) ([]Board, error) {
 			boards = append(boards, Board{
 				Protocol:   NetworkProtocol,
 				Address:    port.GetPort().GetAddress(),
+				Serial:     serial,
```
lucarin91: In network mode you need to get it like the hostname

```
port.GetPort().GetProperties()["serial_number"];
```

lucarin91 (changes requested):

mirkoCrobu:
> > @lucarin91 How can I test it? How App Lab calls this func?
> 
> Probably the best solution is just to write a test or an example that uses that library

To write a test, I would need to mock all the global variables, including "arduinoCLIServer". But in my opinion, it is not very reliable, and I am not sure it is worth it .
What do you think?

PR #84, opened 2025-11-17
Check app status in stop cmd

giulio93:
### Motivation

Before accepting any `arduino-app-cli app stop path:path` we nedd to check that the app explicit in a path is actually running.

### Example Output

`Progress[]: 100%
✓ App "Blink LED" started successfully 
arduino@OOOSUSANNA:/$ arduino-app-cli app stop examples:blink-with-ui 
App "Blink LED with UI" is not running
arduino@OOOSUSANNA:/$ `

### DIFF
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -388,7 +388,7 @@ func stopAppWithCmd(ctx context.Context, docker command.Cli, app app.ArduinoApp,
 			return
 		}
 		if appStatus.Status != StatusStarting && appStatus.Status != StatusRunning {
-			yield(StreamMessage{error: fmt.Errorf("App %q is not running", app.Name)})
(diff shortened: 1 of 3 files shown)


internal/orchestrator/orchestrator.go
```diff
@@ -377,11 +377,21 @@ func getVideoDevices() map[int]string {
...
 
+		appStatus, err := getAppStatus(ctx, docker, app)
+		if err != nil {
+			yield(StreamMessage{error: err})
+			return
+		}
+		if appStatus.Status != StatusStarting && appStatus.Status != StatusRunning {
+			yield(StreamMessage{error: fmt.Errorf("app %q is not running", app.Name)})
+			return
+		}
```
lucarin91: You should use `getRunningApp`

PR #106, opened 2025-11-25
feat: Do not exit from default Python script

per1234:
### Motivation

When a new App is created, it is populated with a simple Python script and "bare minimum" Arduino sketch.

For initial explorations of a new system, or when troubleshooting, it is common to create very simple programs (AKA "Hello, world!"). Although a complex App will typically consist of a Python script and Arduino sketch program working in coordination, for users with prior experience with Arduino the natural approach to creating a minimal App will be to simply write some familiar Arduino sketch code, leaving the default Python script code as-is.

The App is considered to be in a "stopped" state as soon as the Python script exits. This is the correct approach, but may be confusing to users due to the fact that, except perhaps under exceptional conditions, the Arduino sketch program runs perpetually. The author of a minimal sketch-based "Hello, world!" App will find it unintuitive if their App goes into a "stopped" state immediately after starting. The previous default Python script produced exactly that result.

### Change description

The default Python script is hereby changed to the more intuitive behavior of running perpetually.

### Additional Notes

Arduino App Lab 0.2.0 treated the App as being in a running state until the user explicitly stopped it. So even though Arduino App CLI considered the App with the previous Python code to have stopped, App Lab provided the users with the expected behavior where they would continue to be able to see the output from the sketch program in the App Lab Serial Monitor.

This behavior changed in Arduino App Lab 0.2.4. It now considers the App to be stopped when Arduino App CLI considers it stopped. I believe the new behavior of Arduino App Lab is correct, but the users who are accustomed to their Apps having the previous behavior are interpreting this as a bug:

- https://forum.arduino.cc/t/app-lab-monitor-not-showing/1416282
- https://forum.arduino.cc/t/version-0-2-4-won-t-run-apps/1416301
- https://forum.arduino.cc/t/what-works-using-just-ide-2-and-the-stm32/1410992/20 (thanks @mjs513 🙂) 

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/app/generator/app_template/python/main.py
+++ b/internal/orchestrator/app/generator/app_template/python/main.py
@@ -1,6 +1,15 @@
-def main():
-    print("Hello World!")
+import time
+
+from arduino.app_utils import App
+
+print("Hello world!")
 
 
-if __name__ == "__main__":
-    main()
+def loop():
+    """This function is called repeatedly by the App framework."""
+    # You can replace this with any code you want your App to run repeatedly.
+    time.sleep(10)
+
+
+# See: https://docs.arduino.cc/software/app-lab/tutorials/getting-started/#app-run
+App.run(user_loop=loop)
--- a/internal/orchestrator/app/generator/testdata/app-all.golden/python/main.py
+++ b/internal/orchestrator/app/generator/testdata/app-all.golden/python/main.py
@@ -1,6 +1,15 @@
-def main():
-    print("Hello World!")
+import time
+
+from arduino.app_utils import App
+
+print("Hello world!")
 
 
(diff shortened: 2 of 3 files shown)


internal/orchestrator/app/generator/app_template/python/main.py
```diff
@@ -1,5 +1,8 @@
 def main():
```
dido18:
```suggestion
import time 

def main():
```

robgee86:
I like the proposal, I'm adding an observation: in the examples we sponsor using the `App.run()` approach that achieves the same objective.
For consistency, we could use that instead, and the default Python script would become:

```python
from arduino.app_utils import App

print("Hello world!")

App.run()
```

PR #116, opened 2025-11-27
fix: make app failed if a container is stopped

giulio93:
### Motivation

If at least a container of an application is stopped due to any reason, the application status should be failed.
In this way even if other containers are up and running the state of the application state is set as failed, and the user can debug the failed container.

### DIFF
--- a/internal/orchestrator/status.go
+++ b/internal/orchestrator/status.go
@@ -46,13 +46,12 @@ func StatusFromDockerState(s container.ContainerState, statusMessage string) Sta
 		return StatusStopped
 	case container.StateExited:
 		if !isExitBySignal(statusMessage) {
-            // The app exited on its own, which we consider a failure.
+			// The app exited on its own, which we consider a failure.
 			return StatusFailed
 		}
 		return StatusStopped
 	case container.StateDead:
 		return StatusFailed
-
 	default:
 		panic("unreachable")
 	}


internal/orchestrator/helpers.go
```diff
@@ -93,6 +94,10 @@ func parseAppStatus(containers []container.Summary) []AppStatusInfo {
 			appendResult(appPath, StatusStopping)
 			continue
 		}
+		if slices.ContainsFunc(s, func(v Status) bool { return v == StatusStopped }) {
```
lucarin91: we should have some test for checking the whole app transicition, I want to avoid that if a stop an application this transitioning in the fail state. So this check is ok, if all the other container are running, but if I am stopping an app, I don't want to get a fail state during the transition from running to stopped

internal/orchestrator/helpers.go
```diff
@@ -73,27 +90,33 @@ func parseAppStatus(containers []container.Summary) []AppStatusInfo {
...
 
 		// ...else we have multiple different status we calculate the status
 		// among the possible left: {failed, stopping, starting}
-		if slices.ContainsFunc(s, func(v Status) bool { return v == StatusFailed }) {
+		if slices.ContainsFunc(s, func(v containerStateInfo) bool { return v.State == StatusFailed }) {
+			appendResult(appPath, StatusFailed)
+			continue
+		}
+		if slices.ContainsFunc(s, func(v containerStateInfo) bool {
+			return v.State == StatusStopped && strings.Contains(v.StatusMessage, "Exited (0)")
```
lucarin91: We should account for all exit codes that aren't signals, so everything that is less than 128

If you write an app with Python-like

```python
exit(1)
```

You say that this isn't a fail, which in fact it is, because the app exits right after be started.

internal/orchestrator/helpers.go
```diff
@@ -39,6 +39,10 @@ type AppStatusInfo struct {
 	Status  Status
 }
 
+type containerState struct {
+	Status Status
+}
+
```
lucarin91: I think this is a leftover. Can we remove it?

```suggestion
```

PR #128, opened 2025-12-02
fix: force error on apt update

lucarin91:
### Motivation

By default, apt-get update doesn't retunrs any error in the status code if something went wrong during repository update. We want instead to catch this error, which usually means that the board wasn't able to validate the apt repo certificates.

### Change description

Add a config file to enable this behavior.

### Reviewer checklist


lucarin91:
close in favour of https://github.com/arduino/arduino-app-cli/pull/142

PR #143, opened 2025-12-04
feat: do not return hidden variables

dido18:
### Motivation
Some brick variables must not be shown via the AppLab since they are "internal" variables of the bricks.

An example of such variables is `CUSTOM_MODEL_PATH`.

### Change description
Support the `hidden` boolean variable added in the brick index to hide variables.
- do not return the variables in the `GET /v1/bricks/{id}` and `GET /v1/apps/{id}/bricks/{id}` with `hidden:true` 
- A hidden variable could be added "manually" to the `app.yaml`

### Additional Notes

Related pr - https://github.com/arduino/app-bricks-py/pull/38

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/bricksindex/testdata/bricks-list.yaml
+++ b/internal/orchestrator/bricksindex/testdata/bricks-list.yaml
@@ -1,167 +1,158 @@
 bricks:
-  - id: arduino:image_classification
-    name: Image Classification
-    description:
-      "Brick for image classification using a pre-trained model. It processes\
-      \ images and returns the predicted class label and confidence score.\nBrick is\
-      \ designed to work with pre-trained models provided by framework or with custom\
-      \ image classification models trained on Edge Impulse platform. \n"
-    require_container: true
-    require_model: true
-    ports: []
-    model_name: mobilenet-image-classification
-    variables:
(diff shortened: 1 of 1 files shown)


internal/orchestrator/bricks/bricks.go
```diff
@@ -141,6 +141,9 @@ func getInstanceBrickConfigVariableDetails(
 	variableDetails := make([]BrickConfigVariable, 0, len(brick.Variables))
 
 	for _, v := range brick.Variables {
+		if v.Hidden {
```
lucarin91: I would prefer to hide the variable only in the handle return object. IMHO this internal method should be able to get all the variables declared on a brick

lucarin91 (changes requested):

internal/orchestrator/bricks/bricks.go
dido18: I agree, but unfortunately, currenlty the internal structs are used in the HTTP Handler. 
In order to implement this suggestion, I need to create a separate structure in the handlers and map the internal objects.

I would open a ticket for this refactoring, which is for all the code.
WDYT?

PR #146, opened 2025-12-05
Issue 118 move examples

### LINKED ISSUE #118: Move examples (and other data) out of the Arduino home folder
Currently the examples (and other static data) are stored inside the `arduino` home folder.
Current location: `~/.local/share/arduino-app-cli`

- This should also include moving the `~/.arduino15` folder contents. In this case we should put a specific config, in the default contents of the home partition, so that `arduino-cli` will look for the folder in a different location.

The .arduino15 folder was not moved during this phase as we need to further evaluate disk space requirements.
Add some Notes:
This folder is set as the default in arduino-cli (this should not change) and can be overridden by a configuration variable.
This configuration should be used by arduino-app-cli (add to the configuration and check for hardcoded paths).

## Reason
This is needed because at some point the arduino-flasher-cli will support keeping user data, by preserving the user partition (where the arduino home folder is mounted). And we cannot keep some static data in the preserved paths, otherwise that data would not match the software version that gets installed in the root partition (which is not preserved).

---
This PR requires the ARDUINO_APP_CLI__ALLOW_ROOT to be removed from the debian image.
The patch, as well as testing screenshots, are attached

martacarbone:
### Motivation

Currently the examples (and other static data) are stored inside the arduino home folder.
Current location: ~/.local/share/arduino-app-cli

This is needed because at some point the arduino-flasher-cli will support keeping user data, by preserving the user partition (where the arduino home folder is mounted). And we cannot keep some static data in the preserved paths, otherwise that data would not match the software version that gets installed in the root partition (which is not preserved).

### Change description
```
/home/arduino/.local/share/arduino-app-cli/assets -> 
/home/arduino/.local/share/arduino-app-cli/examples -> /var/lib/arduino-app-cli/examples
/home/arduino/.local/share/arduino-app-cli/bootloader_burned.flag -> /var/lib/arduino-app-cli/bootloader_burned.flag
/home/arduino/.local/share/arduino-app-cli/properties.msgpack -> /var/lib/arduino-app-cli/properties.msgpack
```

### Additional Notes
dpkg --install handle obsolete .local file deletion for upgrade from a version lower than the current on (0.7.0)
dpkg --remove delete the system files but the configuration files
dpkg --purge remove all packages related files from the system

#### Test #1
The package status is purged (no file or configurations present on the system)
dpkg --install arduino-app-cli-0.8.0-1_arm64.deb
This is a fresh installation, no old files will be deleted because it is supposed they do no exists.
After the istallation examples are in /var/lib/arduino-app-cli and the application
configuration dir is created in /home/arduino/.config/arduino-app-cli

#### Test #2
The 0.8.0 package is installed over a previous 0.7.0 installation.
This is an upgrade, old files /home/arduino/.local will be deleted.
```dpkg --install arduino-app-cli-0.8.0-1_arm64.deb```
After the installation we should have the same results of Test#1

#### Test #3
dpkg --remove arduino-app-cli
All removed but the configuration file in /home/arduino/.config/arduino-app-cli

#### Test #4
dpkg --purge arduino-app-cli
All removed, no more trace of the package files in the system.

### Reviewer checklist


lucarin91:
I would also move the `properties.msgpack` and the lock file in `/var/lib/arduino-app-cli`.

lucarin91 (changes requested):
I would split this PR into two:
1. move `.cache` and `data` folders somewhere else
2. move `.local/share/` stuff somewhere else

lucarin91:
> I would split this PR into two:
> 
> 1. move `.cache` and `data` folders somewhere else
> 2. move `.local/share/` stuff somewhere else

After some internal discussion, we decided to avoid 1. for now and only move `.local/share` into `/var/`.

PR #153, opened 2025-12-10
Pin the major platform/core version for updates

mirkoCrobu:
### Motivation
closes [738](https://github.com/bcmi-labs/orchestrator/issues/738)

The logic that updates the platform should make sure we do not upgrade above the 1.x.x release.
In other words there should be a variable specifying the maximum major version that will be reported as an update.

This is to avoid auto-updating to a breaking change.>

### Reviewer checklist

### DIFF
--- a/internal/update/arduino/arduino.go
+++ b/internal/update/arduino/arduino.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"log/slog"
 	"slices"
-	"strings"
 	"sync"
 
 	"github.com/arduino/arduino-cli/commands"
@@ -252,12 +251,6 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, packages [
 	targetVersion := pkg.ToVersion
 	if targetVersion == "" {
 		return fmt.Errorf("target version is empty for package '%s'", pkg.Name)
-	}
-	name := pkg.Name
-
-	parts := strings.Split(name, ":")
-	if len(parts) != 2 {
-		return fmt.Errorf("invalid package name %s", name)
 	}
 
 	if err := srv.PlatformInstall(
--- a/internal/update/arduino/arduino_test.go
+++ b/internal/update/arduino/arduino_test.go
@@ -104,6 +104,14 @@ func TestSelectBestVersion(t *testing.T) {
 			expectedVer: "",
 			expectNil:   true,
 		},
+		{
+			name:        "Sorts RC versions correctly",
+			available:   []string{"1.1.0", "1.2.0-rc.3", "2.2.0-rc.4"},
+			installed:   "1.0.0",
+			constraint:  "<2.0.0",
+			expectedVer: "1.2.0-rc.3",
+			expectNil:   false,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


internal/orchestrator/config/config.go
```diff
@@ -105,17 +106,25 @@ func NewFromEnv() (Configuration, error) {
 	if err != nil {
 		return Configuration{}, fmt.Errorf("invalid LIBRARIES_API_URL: %w", err)
 	}
+	maxVersionStr := os.Getenv("ARDUINO_APP_CLI__MAX_UPDATE_MAJOR_VERSION")
```
Xayton: Would be nice to have a comment explaining how this variable works. In particular what is the behavior when the value is zero. It should be "limit to the currently installed major version" right?

internal/update/arduino/arduino.go
```diff
@@ -120,21 +122,83 @@ func (a *ArduinoPlatformUpdater) ListUpgradablePackages(ctx context.Context, _ f
...
 	return []update.UpgradablePackage{{
 		Type:        update.Arduino,
 		Name:        "arduino:zephyr",
 		FromVersion: platformSummary.GetInstalledVersion(),
-		ToVersion:   platformSummary.GetLatestVersion(),
+		ToVersion:   bestVersion,
 	}}, nil
 }
 
+func findBestCandidate(installedStr string, availableVersions []string, constraint semver.Constraint) (string, error) {
```
lucarin91: based on the comment above I think you can remove this

mirkoCrobu: done: 388b275338f37d7bf2e2cd7673a9fd9f2ec4058e

internal/update/arduino/arduino.go
```diff
@@ -121,20 +124,66 @@ func (a *ArduinoPlatformUpdater) ListUpgradablePackages(ctx context.Context, _ f
...
+		if len(parts) > 0 {
+			major, err := strconv.Atoi(parts[0])
+			if err == nil {
+				newConstraintStr := fmt.Sprintf("<%d.0.0", major+1)
+				if newC, err := semver.ParseConstraint(newConstraintStr); err == nil {
+					constraint = newC
+				}
+			}
+		}
 	}
```
lucarin91: Why do we need this check?

lucarin91 (changes requested):

internal/update/arduino/arduino.go
mirkoCrobu: This check covers the following scenario:
- Initial constraint is <1.0.0
- Current installed version is 1.2.3

In this case, we adapt the constraint to <2.0.0. Without this adjustment, valid upgrades (e.g., to 1.3.0) would be ignored because they don't match the initial strict constraint.

internal/update/arduino/arduino.go
```diff
@@ -185,49 +234,38 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, names []st
...
-			Architecture:     "zephyr",
-			SkipPostInstall:  false,
-			SkipPreUninstall: false,
+	targetVersion := packages[0].ToVersion
+	name := packages[0].Name
+
+	if targetVersion == "" {
+		if len(packages) > 0 {
+			return fmt.Errorf("no package of type '%s' found in the upgrade request", name)
+		}
```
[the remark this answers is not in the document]
mirkoCrobu: done

internal/orchestrator/config/config.go
```diff
@@ -106,6 +109,14 @@ func NewFromEnv() (Configuration, error) {
 		return Configuration{}, fmt.Errorf("invalid LIBRARIES_API_URL: %w", err)
 	}
 
+	constraintStr := cmp.Or(os.Getenv("ARDUINO_APP_CLI__UPDATE_VERSION_CONSTRAINT"), "<1.0.0")
```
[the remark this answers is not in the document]
mirkoCrobu: done

internal/update/arduino/arduino.go
```diff
@@ -121,20 +125,66 @@ func (a *ArduinoPlatformUpdater) ListUpgradablePackages(ctx context.Context, _ f
...
+	}
+
+	bestVersion := selectBestVersion(availableReleases, installedV, a.constraint)
+
+	if bestVersion == nil {
+		return nil, nil
+	}
+
+	if bestVersion.Equal(installedV) {
+		return nil, nil
```
dido18: also here empty slice ?

internal/update/arduino/arduino.go
```diff
@@ -185,49 +235,43 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, names []st
...
-			Instance:         inst,
-			PlatformPackage:  "arduino",
-			Architecture:     "zephyr",
-			SkipPostInstall:  false,
-			SkipPreUninstall: false,
+
+	if len(packages) == 0 {
+		return nil
+	}
+
```
dido18: len(packages) is always 1, and it contains the `Name:      "arduino:zephyr",` (see https://github.com/arduino/arduino-app-cli/blob/feeef7e52876c40b80532fba015b91e82ae0770a/internal/update/arduino/arduino.go#L128)

I would add a check on this to make it clear that the function upgrade only one package "ardiuno:zephyr"

IMHO the interface is not needed here and  it could be refactored in a second PR

internal/update/arduino/arduino.go
```diff
@@ -185,49 +235,43 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, names []st
...
+	if targetVersion == "" {
+		if len(packages) > 0 {
+			return fmt.Errorf("no package of type '%s' found in the upgrade request", name)
+		}
+		return fmt.Errorf("package list is empty")
+	}
+
+	parts := strings.Split(name, ":")
+	if len(parts) != 2 {
+		return fmt.Errorf("invalid package name")
```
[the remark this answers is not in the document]
mirkoCrobu: done

dido18 (approved):
LGTM
Some minor changes requested

internal/update/arduino/arduino.go
```diff
@@ -185,49 +235,41 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, names []st
...
+	targetVersion := pkg.ToVersion
+	if targetVersion == "" {
+		return fmt.Errorf("target version is empty for package '%s'", pkg.Name)
+	}
+	name := pkg.Name
+
+	parts := strings.Split(name, ":")
+	if len(parts) != 2 {
+		return fmt.Errorf("invalid package name %s", name)
+	}
```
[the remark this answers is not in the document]
mirkoCrobu: done

PR #157, opened 2025-12-15
feat(pkg/boards): get on board image version

mirkoCrobu:
### Motivation

closes [#41 ](https://github.com/arduino/arduino-flasher-cli/issues/41)
Get the version of the image on the board to know if it supports user preservation when it is not in EDL mode. The equivalent of adb shell 'cat /etc/buildinfo' could be sufficient, but we want to return only the BUILD_ID.
If /etc/buildinfo does not exist, assume it is an R0 image and return it as the default value.

At this level, there is no way to check for EDL mode.
The current behavior is "if the board is detected. It is not in EDL mode."
In any case, this check will be done by the caller.

Change GetOsImageVersion()
Add test if possible
log - INFO

manual test, add the call in main() log - ERROR
### Reviewer checklist

### DIFF
--- a/pkg/board/os_image_test.go
+++ b/pkg/board/os_image_test.go
@@ -77,7 +77,7 @@ func TestParseOSImageVersion(t *testing.T) {
 	}{
 		{
 			name:     "valid build id",
-			input:    "BUILD_ID=\"20251006-395\"\nVARIANT_ID=xfce",
+			input:    "BUILD_ID=20251006-395\nVARIANT_ID=xfce",
 			expected: "20251006-395",
 			found:    true,
 		},
--- a/pkg/board/os_image.go
+++ b/pkg/board/os_image.go
@@ -53,7 +53,7 @@ func parseOSImageVersion(r io.Reader) (string, bool) {
 			continue
 		}
 
-		version := strings.Trim(value, `"' `)
+		version := strings.TrimSpace(value)
 		if version != "" {
 			return version, true
 		}


pkg/flasherapi/flasherapi.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if version, ok := ParseOSImageVersion(string(output)); ok {
+		slog.Info("find OS Image version", "version", version)
+		return version, nil
+	}
+	slog.Info("Unable to find OS Image version", "using default version", defaultVersion)
+	return defaultVersion, nil
+}
+
+func ParseOSImageVersion(buildInfo string) (string, bool) {
```
lucarin91:
```suggestion
func parseOSImageVersion(buildInfo string) (string, bool) {
```

pkg/flasherapi/flasherapi.go
```diff
@@ -0,0 +1,118 @@
...
+	output, err := conn.GetCmd("cat /etc/buildinfo").Output(ctx)
+	if err != nil {
+		return defaultVersion, err
+	}
+
+	if version, ok := ParseOSImageVersion(string(output)); ok {
+		slog.Info("find OS Image version", "version", version)
+		return version, nil
+	}
+	slog.Info("Unable to find OS Image version", "using default version", defaultVersion)
```
lucarin91:
```suggestion
	slog.Warn("Unable to find OS Image version", "using default version", defaultVersion)
```

pkg/flasherapi/flasherapi.go
```diff
@@ -27,34 +29,42 @@ const R0_IMAGE_VERSION_ID = "20250807-136"
...
-	slog.Info("Unable to find OS Image version", "using default version", R0_IMAGE_VERSION_ID)
-	return R0_IMAGE_VERSION_ID, nil
+	slog.Warn("Unable to find OS Image version", "using_default", R0_IMAGE_VERSION_ID)
+	return R0_IMAGE_VERSION_ID
 }
 
-func ParseOSImageVersion(buildInfo string) (string, bool) {
-	for _, line := range strings.Split(buildInfo, "\n") {
-		line = strings.TrimSpace(line)
+func parseOSImageVersion(r io.Reader) (string, bool) {
```
lucarin91: Can you add a test for this function?

pkg/flasherapi/flasherapi.go
```diff
@@ -0,0 +1,69 @@
...
+	"strings"
+
+	"github.com/arduino/arduino-app-cli/pkg/board/remote"
+)
+
+const R0_IMAGE_VERSION_ID = "20250807-136"
+
+// GetOSImageVersion returns the version of the OS image used in the board.
+// It is used by the AppLab to enforce image version compatibility.
+func GetOSImageVersion(ctx context.Context, conn remote.RemoteConn) string {
```
lucarin91:
```suggestion
func GetOSImageVersion(conn remote.RemoteConn) string {
```

pkg/flasherapi/flasherapi.go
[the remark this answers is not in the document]
mirkoCrobu: yes, I agree

pkg/board/os_image_test.go
```diff
@@ -0,0 +1,111 @@
...
+func TestParseOSImageVersion(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+		found    bool
+	}{
+		{
+			name:     "valid build id",
+			input:    "BUILD_ID=\"20251006-395\"\nVARIANT_ID=xfce",
```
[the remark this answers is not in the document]
mirkoCrobu: right

lucarin91 (approved):

PR #161, opened 2025-12-16
fix(update): always try to restart the app-cli daemon

lucarin91:
### Motivation

We want to always try to restart the `arduino-app-cli` daemon, no matter what happened.

### Reviewer checklist

### DIFF
--- a/internal/update/apt/service.go
+++ b/internal/update/apt/service.go
@@ -83,7 +83,8 @@ func (s *Service) UpgradePackages(ctx context.Context, names []string) (<-chan u
 		defer s.lock.Unlock()
 		defer close(eventsCh)
 
-		// We try anyway to restart the service.
+		// At the end of the upgrade, always try to restart the services (that need it).
+		// This makes sure key services are restarted even if an error happens in the upgrade steps (for examples container images download).
 		defer func() {
 			eventsCh <- update.NewDataEvent(update.RestartEvent, "Upgrade completed. Restarting ...")
 


lucarin91:
stated the update test here https://github.com/arduino/arduino-app-cli/actions/runs/20262748460

internal/update/apt/service.go
```diff
@@ -83,6 +83,17 @@ func (s *Service) UpgradePackages(ctx context.Context, names []string) (<-chan u
 		defer s.lock.Unlock()
 		defer close(eventsCh)
 
+		// We try anyway to restart the service.
```
Xayton:
```suggestion
		// At the end of the upgrade, always try to restart the services (that need it).
		// This makes sure key services are restarted even if an error happens in the upgrade steps (for examples container images download).
```

internal/update/apt/service.go
```diff
@@ -83,6 +83,18 @@ func (s *Service) UpgradePackages(ctx context.Context, names []string) (<-chan u
...
+		// This makes sure key services are restarted even if an error happens in the upgrade steps (for examples container images download).
+		defer func() {
+			eventsCh <- update.NewDataEvent(update.RestartEvent, "Upgrade completed. Restarting ...")
+
+			err := restartServices(ctx)
+			if err != nil {
+				eventsCh <- update.NewErrorEvent(fmt.Errorf("error restarting services after upgrade: %w", err))
+				return
+			}
+		}()
```
dido18: We need to carefully check how Applab is interpreting the `restarted` and `error` event.

For example, what happens if there is an error (for example, when the runupgradeCommand is executed)?
We send the following SSE events:
 - error                                                => mark as failed the update 
 - restarted (due to defer)               => ?? mark as succeeded =

PR #179, opened 2025-12-24
fix(update): don't stop at arduino platform errors

lucarin91:
### Motivation

After this  [PR](https://github.com/arduino/arduino-app-cli/pull/176) the update fails if the platform update fails. We should revert to the old behavior, which is attempting to do both updates, i.e., platform and deb packages.

### Reviewer checklist


lucarin91:
tests run successfully https://github.com/arduino/arduino-app-cli/actions/runs/20485386065

PR #181, opened 2025-12-29
[API] Add "import/export app" commands

mirkoCrobu:
### Change description

We should add a command and API endpoint (to be used by the App Lab), to export and import Arduino Apps.

The exported format should be the app folder structure, zipped as a single file
When exporting, remove .cache, secrets and data (or provide an option to keep or exclude data)
When importing, the structure and contents of the app must be validated first

# Export 
`GET /api/v1/apps/{appId}/export?include_data=true`
### parameters:
 `Include_data `: boolean,  optional,  default: false 

### Responses
**200 OK**
Content-Type: application/zip
Content-Disposition: attachment; filename="my_app.zip"

**400 Bad Request**
```json
{
  "details": "The parameter 'include_data' must be a boolean.",
}
```

**404 Not Found**
```json
{
  "details": "App with ID '12345' not found.",
}
```

**500 Internal Server Error**
```json
{
  "details": "Failed to generate archive.",
}
```

### Export rules
**Always exclude:** 
- .cache directory
- secrets(to check)
- data dir  optional( based on the paramiter)

#### In-Memory Processing (Buffer)
**Pro**
- simple implementation
- zip creation completes quickly
- Content-Length can be calculated
- easy error handling
- ZIP always valid

**Cons**
- RAM consumption depends on app size and number of concurrent requests
- potential GC pressure on board if multiple exports run simultaneously
- may not scale well for larger apps

#### Direct Stream Processing
**Pro**
- constant and low memory usage
- read files once, no temp files needed
- safe for multiple concurrent requests

**Cons**
more complex implementation
error during streaming may result in a partially corrupted ZIP
Content-Length cannot be easily known in advance
may be slower than in-memory buffer if using compression

#### Temporary Zip on Disk
**Pro**
- simple implementation
- Content-Length can be calculated
- low RAM usage
- easy error handling
- ZIP always valid

**Cons**
- requires sufficient disk space
- involves double I/O (write + read) → slower than pure in-memory for small apps

# Import 
`POST /api/v1/apps/import`

Headers
Content-Type: multipart/form-data

**Body Parameters**
file: Binary (The .zip file to upload). Required.

### Responses
201 Created Import successful. Returns the id(`same response as POST /v1/apps`) of the newly created app.

```json
{
  "id": "12345"
}
```

**400 Bad Request** (Validation Errors) The file is invalid, corrupt, or the directory structure is incorrect.
```json
{
  "details": "Missing required file 'app.yaml' ."
}
```

**409 Conflict** An app with the same name already exists. 
```json
{
  "details": "App with ID '12345' already exists."
}
```

**500 Internal Server Error**
```json
{
  "details": "Failed to process the archive."
}
```

### Validation Rules
The server must validate the archive before finalizing the import.
If any check fails, the process aborts, and no files are written to the final app destination.
Verify that the file is a valid ZIP archive and check that the archive must contain essential folders:
- app.yaml (content validation required)
- sketch dir(content validation required)
- python dir(content validation required)

Reuse the existing code for validation  (also check [this]https://github.com/arduino/arduino-app-cli/blob/5deb69a86b07f7831819460475e8addf779d0549/internal/orchestrator/app/validator.go#L11

### Sanitization(after validation): **we just import everithing is a valid zip**
Drop any file that is not one of the following: 
- app.yaml
- python dir
- sketch dir

### Implementation Details
1. Stream to Disk: The incoming stream is saved to a temporary file (e.g., /tmp/upload_my_app.zip).
2. validation step. If Invalid: Delete the temp files and return an error. If Valid: perform sanitization and extract files to the final destination (/ArduinoApps folder).
3. Cleanup the temporary files.
4. return response.

### Point to discuss
- Should support override in case of conflict?  --> **NEVER** (With the current proposal, the user must manually delete the existing one before importing the new one) -> app folder e app name should be equal.  two apps  are identical if app folder names are identical and app.yaml names are equal
- Should we set a size limit for the upload? -->not yet
- Should perform security checks against viruses/malware? 
- Should set a timeout to prevent long uploads?  -->not yet

### TODO 
- check folder and name apps convention: folder name and app.yaml app name must be the same? 
-  

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/archive.go
+++ b/internal/orchestrator/archive.go
@@ -297,18 +297,18 @@ func validateAppZipContent(r *zip.Reader) error {
 	}
 
 	if !hasAppYaml {
-		return errors.New("missing app.yaml")
+		return errors.New(" missing app.yaml")
 	}
 	if !hasMainPy {
-		return errors.New("missing python/main.py")
+		return errors.New(" missing python/main.py")
 	}
 
 	if hasSketchFolder {
 		if !hasSketchIno {
-			return errors.New("sketch folder present but missing .ino file")
+			return errors.New(" sketch folder present but missing .ino file")
 		}
 		if !hasSketchYaml {
-			return errors.New("sketch folder present but missing .yaml file")
+			return errors.New(" sketch folder present but missing .yaml file")
 		}
 	}
 


internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +959,264 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
...
+
+	if err := os.MkdirAll(finalDestPath, 0755); err != nil {
+		return "", fmt.Errorf("unable to create app directory. %w", err)
+	}
+
+	if err := extractZip(&r.Reader, finalDestPath); err != nil {
+		_ = os.RemoveAll(finalDestPath)
+		return "", err
+	}
+
```
dido18: I find this import logic complicated to follow.

My suggestion is:
- unzip the zip into the `~/ArduinoApps/<name>` 
- do basic check (app.yaml exist)

internal/api/handlers/app_import.go
```diff
@@ -29,6 +29,12 @@ func HandleAppImport(
 	idProvider *app.IDProvider,
 ) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		folderName := strings.TrimSpace(r.FormValue("folder_name"))
```
dido18: I suggest calling the `ParseMultipartForm`  explicitely  to check size limits

```
err := r.ParseMultipartForm(32 << 10) // 32 MB
	if err != nil {
		log.Fatal(err)
	}
```

and then parse the json and file 
```
// ---- Validate metadata (JSON) ----
	metaValues := r.MultipartForm.Value["metadata"]
	if len(metaValues) != 1 {
		http.Error(w, "exactly one 'metadata' field is required", http.StatusBadRequest)
		return
	}

	var meta Metadata
	if err := json.Unmarshal([]byte(metaValues[0]), &meta); err != nil {
		http.Error(w, "invalid metadata JSON", http.StatusBadRequest)
		return
	}

	// ---- Validate file ----
	files := r.MultipartForm.File["file"]
	if len(files) != 1 {
		http.Error(w, "exactly one zip file is required", http.StatusBadRequest)
		return
	}
```

internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +961,252 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
 	return app.FullPath.RemoveAll()
 }
 
+func ExportApp(
```
dido18: I would rename the method like "ExportAppZip` since it exports an app as a zip.

internal/api/handlers/app_import.go
```diff
@@ -0,0 +1,88 @@
...
+type ImportResponse struct {
+	ID string `json:"id"`
+}
+
+func HandleAppImport(
+	cfg config.Configuration,
+	idProvider *app.IDProvider,
+) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		folderName := strings.TrimSpace(r.FormValue("folder_name"))
```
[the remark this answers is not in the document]
Xayton: But this is a non-requested feature. It's OK to do, but not if it slows down the release of the feature (or the linked activity of providing CLI commands to import/export).

I think it should not go into this PR.

internal/api/handlers/app_import.go
mirkoCrobu: ParseMultipartForm is called internally by  `r.FormFile("file")` some lines below. 
but if the intent is to avoid space issues we should use: `http.MaxBytesReader`. Should we implement it?
```

File: internal/api/handlers/app_import.go [hunk 5]
```diff
@@ -0,0 +1,88 @@
...
+		return
+	}
+	if errors.Is(err, orchestrator.ErrBadRequest) ||
+		strings.Contains(err.Error(), "not a valid zip file") {
+		render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: err.Error()})
+		return
+	}
+
+	render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "failed to process the archive: " + err.Error()})
+}
```
dido18: I don't think this function is useful. I prfere to put it inline

internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +962,258 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
 	return app.FullPath.RemoveAll()
 }
 
+func ExportAppZip(
```
dido18: I would add some unit tests for this package
For example, given an app, export it and then import it and check for no errors (happy path).

internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +962,258 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
...
+}
+
+func ImportAppFromZip(
+	cfg config.Configuration,
+	zipPath string,
+	folderName string,
+	idProvider *app.IDProvider,
+) (string, error) {
+	appsBasePath := cfg.AppsDir()
+	appFolderName, err := sanitizeAndValidateFolderName(folderName)
```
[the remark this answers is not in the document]
mirkoCrobu: done

internal/api/handlers/app_import.go
mirkoCrobu: done

internal/api/handlers/app_import.go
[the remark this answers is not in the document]
mirkoCrobu: done

internal/api/docs/openapi.yaml
```diff
@@ -522,6 +522,35 @@ paths:
...
+        required: true
+        schema:
+          description: application identifier.
+          type: string
+      responses:
+        "200":
+          content:
+            application/zip:
+              schema:
+                format: base64
```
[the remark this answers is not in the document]
mirkoCrobu: You are right, well spotted. We have base64 because the generator cannot interpret DataStructure: []byte{} as binary automatically, since DataStructure is an interface.  we can fix this with the following solution:
1852e2c99b06a6c155ab6c7d39f6f8e7c11502eb

internal/api/handlers/app_export.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/arduino/arduino-app-cli/internal/orchestrator/config"
+	"github.com/arduino/arduino-app-cli/internal/render"
+)
+
+func HandleAppExport(
+	idProvider *app.IDProvider,
+	cfg config.Configuration,
+) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		id, err := idProvider.IDFromBase64(r.PathValue("appID"))
```
dido18: I would use only `id`.
```suggestion
		id, err := idProvider.IDFromBase64(r.PathValue("id"))
```

internal/api/handlers/app_import.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		if strings.TrimSpace(opts.FolderName) == "" {
+			render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "json field 'folder_name' is required"})
+			return
+		}
+
+		tempFile, err := os.CreateTemp("", "app-import-*.zip")
+		if err != nil {
+			slog.Error("unable to create temp file", slog.String("error", err.Error()))
```
dido18: It should work also like this
```suggestion
			slog.Error("unable to create temp file", "err", err))
```

internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +962,258 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
...
+
+	return buf.Bytes(), nil
+}
+
+func ImportAppFromZip(
+	cfg config.Configuration,
+	zipPath string,
+	folderName string,
+	idProvider *app.IDProvider,
+) (string, error) {
```
dido18: return the app id
```suggestion
) (app.ID, error) {
```

internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +962,258 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
...
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+func ImportAppFromZip(
+	cfg config.Configuration,
+	zipPath string,
+	folderName string,
```
dido18: Remove this and infer the name of the app from the app.yaml

dido18 (changes requested):

internal/orchestrator/app/validator.go
```diff
@@ -43,3 +45,55 @@ func ValidateBricks(a AppDescriptor, index *bricksindex.BricksIndex) error {
 	}
 	return allErrors
 }
+
+func ValidateStructure(filePaths []string) error {
```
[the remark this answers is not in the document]
mirkoCrobu: removed. no more necessary

internal/api/handlers/app_export.go
```diff
@@ -0,0 +1,48 @@
...
+				render.EncodeResponse(w, http.StatusNotFound, models.ErrorResponse{Details: "unable to find the app"})
+			} else {
+				render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "unable to parse the app"})
+			}
+			return
+		}
+
+		zipBytes, fileName, err := orchestrator.ExportAppZip(r.Context(), app)
+		if err != nil {
+			slog.Error("failed to export app", slog.String("app_id", id.String()), slog.String("error", err.Error()))
```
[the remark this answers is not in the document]
mirkoCrobu: done

internal/api/handlers/app_import.go
```diff
@@ -0,0 +1,98 @@
...
+			render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "internal server error"})
+			return
+		}
+		tempPath := tempFile.Name()
+		defer os.Remove(tempPath)
+
+		if _, err := io.Copy(tempFile, file); err != nil {
+			tempFile.Close()
+			slog.Error("unable to save upload to temp file", slog.String("error", err.Error()))
```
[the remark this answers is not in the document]
mirkoCrobu: done

internal/api/handlers/app_export.go
```diff
@@ -0,0 +1,48 @@
...
+		id, err := idProvider.IDFromBase64(r.PathValue("appID"))
+		if err != nil {
+			render.EncodeResponse(w, http.StatusPreconditionFailed, models.ErrorResponse{Details: "invalid id"})
+			return
+		}
+		app, err := app.Load(id.ToPath())
+		if err != nil {
+			slog.Error("Unable to parse the app.yaml", slog.String("error", err.Error()), slog.String("path", id.String()))
+			if errors.Is(err, os.ErrNotExist) {
+				render.EncodeResponse(w, http.StatusNotFound, models.ErrorResponse{Details: "unable to find the app"})
```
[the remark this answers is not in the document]
mirkoCrobu: done

internal/api/handlers/app_import.go
```diff
@@ -42,44 +39,26 @@ func HandleAppImport(
...
 		tempFile, err := os.CreateTemp("", "app-import-*.zip")
 		if err != nil {
-			slog.Error("unable to create temp file", slog.String("error", err.Error()))
+			slog.Error("unable to create temp file", "err", err)
 			render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "internal server error"})
 			return
 		}
-		tempPath := tempFile.Name()
-		defer os.Remove(tempPath)
+		tempFilePath := paths.New(tempFile.Name())
```
dido18: I think paths has some utility to create a temp file. See  `paths.MkTempFile().`

internal/api/handlers/app_import.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/arduino/go-paths-helper"
+
+	"github.com/arduino/arduino-app-cli/internal/api/models"
+	"github.com/arduino/arduino-app-cli/internal/orchestrator"
+	"github.com/arduino/arduino-app-cli/internal/orchestrator/app"
+	"github.com/arduino/arduino-app-cli/internal/orchestrator/config"
+	"github.com/arduino/arduino-app-cli/internal/render"
+)
+
+type ImportResponse struct {
```
dido18:
```suggestion
type AppImportResponse struct {
```

internal/api/handlers/app_import.go
```diff
@@ -0,0 +1,87 @@
...
+				render.EncodeResponse(w, http.StatusConflict, models.ErrorResponse{Details: err.Error()})
+			case errors.Is(err, orchestrator.ErrBadRequest) || strings.Contains(err.Error(), "not a valid zip file"):
+				render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: err.Error()})
+			default:
+				render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "failed to process the archive: " + err.Error()})
+			}
+			return
+		}
+
+		slog.Info("app imported successfully", "app_id", appID)
```
dido18:
```suggestion
```

internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +963,294 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
...
+			return nil
+		}
+		if strings.HasPrefix(relPath, ".cache") {
+			if d.IsDir() {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+
+		if !includeData && strings.HasPrefix(relPath, "data") {
```
dido18: why has prefix and not equalTo ?

internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +963,294 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
...
+			return fmt.Errorf("%w: sketch folder present but missing .yaml file", ErrBadRequest)
+		}
+	}
+
+	return nil
+}
+
+var validFolderName = regexp.MustCompile(`^[a-z0-9_-]{1,50}$`)
+
+func sanitizeAndValidateFolderName(name string) (string, error) {
```
dido18: This seems not to be used anywhere

dido18:
Test on a real board.
- export an app with data folder
```sh
 wget "127.0.0.1:8800/v1/apps/dXNlcjpjb3B5LW9mLXN5c3RlbS1yZXNvdXJjZXMtbG9nZ2Vy/export?include_data=true"
```
- error of app already exist:  
```sh
 curl -X POST http://127.0.0.1:8800/v1/apps/import \
     -F "file=@app.zip"
{"details":"app already exists"
```
- export 
```
 wget "127.0.0.1:8800/v1/apps/dXNlcjpjb3B5LW9mLXN5c3RlbS1yZXNvdXJjZXMtbG9nZ2Vy/export?include_data=true
```
- import the exported zip 
```
curl -X POST http://127.0.0.1:8800/v1/apps/import    \
  -F "file=@app.zip"
{"id":"dXNlcjpjb3B5LW9mLXN5c3RlbS1yZXNvdXJjZXMtbG9nZ2Vy"}
```

- check the imported app structure
```
arduino@dido:~/ArduinoApps/copy-of-system-resources-logger$ tree -L 3
``` sh
├── app.yaml
├── assets
│   ├── app.js
│   ├── docs_assets
│   │   ├── launch-app.png
│   │   └── system-resource-log.png
│   ├── fonts
│   │   ├── fonts.css
│   │   ├── Open Sans
│   │   └── Roboto
│   ├── img
│   │   ├── favicon.png
│   │   ├── info.svg
│   │   ├── nodata.svg
│   │   ├── RGB-Arduino-Logo_Color Inline Loop.svg
│   │   └── RGB-Arduino-Logo_Color Inline Loop.svg.license
│   ├── index.html
│   ├── libs
│   │   └── socket.io.min.js
│   └── style.css
├── data
│   └── influx-data
│       ├── engine
│       ├── influxd.bolt
│       └── influxd.sqlite
├── python
│   ├── main.py
│   └── requirements.txt
└── README.md

```
- start the app again 
```
arduino-app-cli app  start user:copy-of-system-resources-logger
[INFO] Starting app "Copy of System resources logger"
Progress[preparing]: 0%
[INFO] python provisioning
Progress[python provisioning]: 0%
[INFO] python downloading
[INFO]  Container copy-of-system-resources-logger-dbstorage-influx-1  Created
[INFO]  Container copy-of-system-resources-logger-main-1  Created
[INFO]  Container copy-of-system-resources-logger-dbstorage-influx-1  Starting
[INFO]  Container copy-of-system-resources-logger-dbstorage-influx-1  Started
[INFO]  Container copy-of-system-resources-logger-dbstorage-influx-1  Waiting
[INFO]  Container copy-of-system-resources-logger-dbstorage-influx-1  Healthy
[INFO]  Container copy-of-system-resources-logger-main-1  Starting
[INFO]  Container copy-of-system-resources-logger-main-1  Started
Progress[]: 100%
✓ App "Copy of System resources logger" started successfull
```

internal/orchestrator/orchestrator.go
```diff
@@ -956,6 +966,279 @@ func DeleteApp(ctx context.Context, dockerClient command.Cli, app app.ArduinoApp
...
+	if strings.TrimSpace(appDescriptor.Name) == "" {
+		return app.ID{}, fmt.Errorf("%w: app name is missing", ErrBadRequest)
+	}
+
+	finalDestPath, appExists := findAppPathByName(appDescriptor.Name, cfg)
+	if appExists {
+		return app.ID{}, ErrAppAlreadyExists
+	}
+
+	tempDirName := fmt.Sprintf(TmpAppPrefix+"%s", rand.Text())
```
dido18: Could you add a small comment on the WHY we are creating a temp dir?  The main motivation is the atomic swap of the folder

internal/orchestrator/archive.go
```diff
@@ -0,0 +1,315 @@
...
+	if !hasMainPy {
+		return fmt.Errorf(" missing python/main.py")
+	}
+
+	if hasSketchFolder {
+		if !hasIno {
+			return fmt.Errorf(" sketch folder present but missing .ino file")
+		}
+		if !hasSketchYaml {
+			return fmt.Errorf("sketch folder present but missing .yaml file")
```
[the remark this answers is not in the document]
mirkoCrobu: done

internal/orchestrator/archive.go
```diff
@@ -0,0 +1,315 @@
...
+	if !hasAppYaml {
+		return fmt.Errorf(" missing app.yaml")
+	}
+	if !hasMainPy {
+		return fmt.Errorf(" missing python/main.py")
+	}
+
+	if hasSketchFolder {
+		if !hasIno {
+			return fmt.Errorf(" sketch folder present but missing .ino file")
```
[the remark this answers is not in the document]
mirkoCrobu: done

internal/orchestrator/archive.go
```diff
@@ -0,0 +1,315 @@
...
+				hasSketchYaml = true
+			}
+		}
+	}
+
+	if !hasAppYaml {
+		return fmt.Errorf(" missing app.yaml")
+	}
+	if !hasMainPy {
+		return fmt.Errorf(" missing python/main.py")
```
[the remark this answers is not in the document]
mirkoCrobu: done

dido18 (approved):

PR #186, opened 2026-01-15
SketchListLibraries API call returns libraries and dependencies

cmaglie:
### Motivation

An Arduino App might depend directly on some sketch libraries and, in turn, depend on other secondary libraries.
This information is encoded in the sketch.yaml file, and considered when adding and removing libraries.

When returning the list of libraries to the App Lab, this information is lost.

* [x] Add a way to filter the libraries by the "secondary dependency" concept, for example, with a query parameter
* [x] Expose the "secondary dependency" concept in the returned data

### Change description

* Added the `hide_deps` query parameter.
* The `secondary dependency` flag was already populated in the existing code, now it is reported in the API response too.

### Additional Notes

The added unit test revealed some existing inconsistencies:
* The HTTP STATUS returned from LibInstall was 201 instead of 200.
* The URL of LibraryList was wrong in the open-api spec
* The LibraryReleaseID is now fully define in open-api spec.

### Reviewer checklist


internal/orchestrator/sketch_libs_release_id.go
```diff
@@ -27,9 +26,9 @@ import (
 // - name[@version]
 // Version is optional, if not provided, the latest version available will be used.
 type LibraryReleaseID struct {
-	Name         string
-	Version      string
-	IsDependency bool
+	Name         string `json:"name"`
```
dido18: We need the return also the libraryID with the `name@version` syntax (like `MsgPack@0.4.2`).
The id is used by the FE to call other endpoints, like `"PUT /v1/apps/{appID}/sketch/libraries/{libRef}"`

dido18 (changes requested):
This is a breaking change. 
 
Current behaviour:
```
curl 127.0.0.1:8800/v1/apps/dXNlcjpzY3JhdGNoLWFyZHVpbm8tYXBw/sketch/libraries | jq
 
{
  "libraries": [
    "MsgPack@0.4.2",
    "DebugLog@0.8.4",
    "ArxContainer@0.7.0",
    "ArxTypeTraits@0.3.1",
    "Arduino_Modulino@0.6.1",
    "Arduino_HS300x@1.0.0",
    "STM32duino VL53L4CD@1.0.5",
    "STM32duino VL53L4ED@1.0.1",
    "Arduino_LSM6DSOX@1.1.2",
    "Arduino_LPS22HB@1.0.2"
  ]
}
```
New behaviour
```
 curl 127.0.0.1:8800/v1/apps/dXNlcjpzY3JhdGNoLWFyZHVpbm8tYXBw/sketch/libraries | jq
{
  "libraries": [
    {
      "name": "MsgPack",
      "version": "0.4.2",
      "is_dependency": false
    },
    {
      "name": "DebugLog",
      "version": "0.8.4",
      "is_dependency": false
    },
    {
      "name": "ArxContainer",
      "version": "0.7.0",
      "is_dependency": false
    },
    {
      "name": "ArxTypeTraits",
      "version": "0.3.1",
      "is_dependency": false
    },
    {
      "name": "Arduino_Modulino",
      "version": "0.6.1",
      "is_dependency": false
    },
    {
      "name": "Arduino_HS300x",
      "version": "1.0.0",
      "is_dependency": false
    },
    {
      "name": "STM32duino VL53L4CD",
      "version": "1.0.5",
      "is_dependency": false
    },
    {
      "name": "STM32duino VL53L4ED",
      "version": "1.0.1",
      "is_dependency": false
    },
    {
      "name": "Arduino_LSM6DSOX",
      "version": "1.1.2",
      "is_dependency": false
    },
    {
      "name": "Arduino_LPS22HB",
      "version": "1.0.2",
      "is_dependency": false
    }
  ]
}
```

I see these options:
 - add a new field (e.g. 'dependencies`) to v1 with the list of string containing the library dependencies
 - deprecated the 'libraries` field in v1 and add another field with the array of objects.
 - make v2

internal/api/docs/openapi.yaml
```diff
@@ -262,7 +262,7 @@ paths:
       summary: Get app exposed ports
       tags:
       - Application
-  /v1/apps/{appID}/sketch/libraries/:
+  /v1/apps/{appID}/sketch/libraries:
```
mirkoCrobu: the final slash is due to https://github.com/arduino/arduino-app-cli/blob/463d0c5222e4e4d441fb35d32bf3b898e38219c4/cmd/gendoc/docs.go#L1116
should be : 			`Path:        "/v1/apps/{appID}/sketch/libraries/", `

Xayton (approved):
Code is OK. Behavior is OK (it works in App Lab and the dependencies are hidden by default now), while still being retro-compatible.

PR #187, opened 2026-01-15
feat(edgeimpulse): support custom ei-models

dido18:
### Motivation
Add the initial support for reading the edge impulse models from the file system.

### Change description
- Extend the `modelindex` package to load the pre-installed models + eimodels.
- add the `POST v1/models/edge-impulse` to install a model

### Additional Notes

How to test on a real board
- Create a `JWT token` using https://docs.edgeimpulse.com/apis/studio/login/get-jwt-token . Insert your username and password to get the token. NOTE: enable the password in the edgeimpulse account if only SSO inabled.
- Copy the JWT token

### Reviewer checklist

### DIFF
--- a/internal/api/handlers/ai_models.go
+++ b/internal/api/handlers/ai_models.go
@@ -35,7 +35,7 @@ import (
 type InstallEIModelRequest struct {
 	ProjectID  int    `json:"project_id" description:"Edge Impulse project ID" example:"123456" required:"true"`
 	ImpulseID  int    `json:"impulse_id" description:"Edge Impulse impulse ID" example:"1" required:"true"`
-	Token      string `json:"token" description:"Edge Impulse API token" example:"your_edge_impulse_api_token" required:"true"`
+	PrjApiKey  string `json:"prj_api_key" description:"Edge Impulse API token" example:"your_edge_impulse_api_token" required:"true"`
 	ModelType  string `json:"model_type" description:"Type of model to build (e.g., 'object-detection')" example:"object-detection" required:"true"`
 	Engine     string `json:"engine" description:"Model engine (e.g., 'tensorflow-lite')" example:"tensorflow-lite" required:"true"`
 	DeviceType string `json:"device_type" description:"Device type for deployment (e.g., 'arduino-uno-q')" example:"arduino-uno-q" required:"true"`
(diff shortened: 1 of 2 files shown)


cmd/arduino-app-cli/internal/servicelocator/servicelocator.go
```diff
@@ -46,7 +47,7 @@ var (
 	})
 
 	GetModelsIndex = sync.OnceValue(func() *modelsindex.ModelsIndex {
-		return f.Must(modelsindex.Load(GetStaticStore().GetAssetsFolder()))
+		return f.Must(modelsindex.Load(GetStaticStore().GetAssetsFolder(), edgeimpulse.New(globalConfig.CustomEiModelsDir())))
```
lucarin91: Should this be called edgeimpulse? Cannot be generic?

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -84,7 +94,7 @@ func (m *ModelsIndex) GetModelsByBrick(brick string) []AIModel {
 
 func (m *ModelsIndex) GetModelsByBricks(bricks []string) []AIModel {
 	var matchingModels []AIModel
-	for _, model := range m.Models {
+	for _, model := range m.GetModels() {
```
giulio93: this is misleading, is strange to have a m.Models internal structure loaded once, and then a function to henhance this structure with the custom models that can change based on a third party.
We should at least rename m.Model to something like m.ArduinoModels, and then  GetCustoModels()
or get the structure to be uptodate every time a model is installed

dido18: see https://github.com/arduino/arduino-app-cli/pull/187/commits/4ca08cd689e8c5c107aa85bba7d22c3ce558540f

dido18:
> @dido18 @giulio93 What do you think of these alternatives using iterators #191? The main objective is to generalize that models are stored in multiple places, and the only things the model index needs to know are a list of iterable objects. Those object will hide their implementation and details. Performance-wise, I would expect that sometimes we can spare some disk access if we find models in the preinstalled interable.

I agree, but for now I would leave as is and generalise in a different PR

internal/orchestrator/modelsindex/client.gen.go
lucarin91: I would definitely avoid committing this file. Can we extract just the part that we need instead of everything?

dido18:
Closed infavour of https://github.com/arduino/arduino-app-cli/pull/214

PR #192, opened 2026-01-19
feat(edgeimpulse):  (list, install)

giulio93:
### Additional Notes

https://github.com/arduino/private-tooling-team/issues/392
### Reviewer checklist


internal/orchestrator/modelsindex/ei_client.go
```diff
@@ -0,0 +1,278 @@
...
+	LastAccessed time.Time `json:"lastAccessed"`
+	LastModified time.Time `json:"lastModified"`
+}
+
+type ProjectResponse struct {
+	Project Project `json:"project"`
+}
+
+func NewEIClient(apiKey string, ApiUrl string) *EIClient {
+	return &EIClient{APIKey: apiKey, ApiUrl: ApiUrl}
```
dido18: Is the apikey passed every time the download of the model is performed ?

internal/orchestrator/modelsindex/ei_client.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func (c *EIClient) DownloadAndInstallModel(ctx context.Context, modelPath string, projectID int, impulseID int) error {
```
dido18: If every time the api key is passed, I would refacotr the code as follows:

```
DownloadAndInstallModel(ctx context.Context, url, apiKey, projectID int, impulseID int, modelPath) error 
```

This function has the logic insides to call the API, check job status,  and download the model inside the path.

The usage is

PR #200, opened 2026-01-25
Add test for custom models variables in app.yaml.

martacarbone:
### Motivation

When a custom model id added to a brick installed of an app, the app.yaml must be edited and the model added to the brick. This is already done by the code, this patch add a test to enforce this behavior.

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/bricks/bricks_test.go
+++ b/internal/orchestrator/bricks/bricks_test.go
@@ -248,13 +248,16 @@ func TestUpdateBrick(t *testing.T) {
 		require.Nil(t, paths.New("testdata/dummy-app-for-model").CopyDirTo(tempDummyApp))
 		bricksIndex, err := bricksindex.Load(paths.New("testdata"))
 		require.Nil(t, err)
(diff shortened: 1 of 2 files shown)


internal/orchestrator/bricks/bricks_test.go
```diff
@@ -242,6 +242,36 @@ func TestUpdateBrick(t *testing.T) {
...
+		require.Nil(t, err)
+		brickService := NewService(nil, bricksIndex, nil)
+
+		modelPath := "/home/arduino/.arduino-bricks/ei-model-123-1/model.eim"
+		modelId := "ei-model-123-1"
+		brickId := "arduino:brick-with-custom-model"
+		req := BrickCreateUpdateRequest{
+			ID: brickId,
+			Variables: map[string]string{
+				"EI_OBJ_DETECTION_MODEL": modelId,
```
dido18:
```suggestion
		req := BrickCreateUpdateRequest{
			ID: brickId,
			Model: modelId,
			Variables: map[string]string{
				"EI_OBJ_DETECTION_MODEL": modelId,
```

PR #214, opened 2026-01-28
Download and Install an EI model

giulio93:
### Motivation

This client allow to build a model on Edge Impulse and download directly on the board using ` .eim` format.
A model.yml file is created in the same folder with the name `ei-model-[projectID]-[modelID]`, this file will report model information that will be used to be associated with the right brick where this model can be used.

### DIFF
--- a/internal/api/docs/openapi.yaml
+++ b/internal/api/docs/openapi.yaml
@@ -970,11 +970,28 @@ paths:
       summary: Get AI model details
       tags:
       - AIModels
-  /v1/models/edge-impulse:
-    post:
+  /v1/models/ei/projects/{projectID}:
+    put:
       description: Download and install a custom Edge Impulse AI model using the provided
         project API key.
       operationId: installEIModel
+      parameters:
+      - description: Edge Impulse project ID
+        in: path
+        name: projectID
+        required: true
+        schema:
+          description: Edge Impulse project ID
+          example: 123456
+          type: integer
+      - description: Edge Impulse project API key
+        in: header
+        name: x-api-key
+        required: true
+        schema:
+          description: Edge Impulse project API key
+          example: your_edge_impulse_api_token
+          type: string
       requestBody:
         content:
           application/json:
@@ -984,18 +1001,8 @@ paths:
                   description: Edge Impulse impulse ID
                   example: 1
                   type: integer
-                prj_api_key:
-                  description: Edge Impulse project API key
-                  example: your_edge_impulse_api_token
-                  type: string
-                project_id:
-                  description: Edge Impulse project ID
-                  example: 123456
-                  type: integer
               required:
-              - project_id
               - impulse_id
-              - prj_api_key
               type: object
       responses:
         "200":
--- a/internal/api/handlers/ai_models.go
+++ b/internal/api/handlers/ai_models.go
@@ -37,9 +37,7 @@ import (
 )
 
 type InstallEIModelRequest struct {
-	ProjectID *int   `json:"project_id" description:"Edge Impulse project ID" example:"123456" required:"true"`
-	ImpulseID *int   `json:"impulse_id" description:"Edge Impulse impulse ID" example:"1" required:"true"`
-	PrjApiKey string `json:"prj_api_key" description:"Edge Impulse API key" example:"your_edge_impulse_api" required:"true"`
+	ImpulseID *int `json:"impulse_id" description:"Edge Impulse impulse ID" example:"1" required:"true"`
 }
 
 func HandleModelsList(modelsIndex *modelsindex.ModelsIndex) http.HandlerFunc {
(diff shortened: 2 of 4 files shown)


dido18 (changes requested):

internal/orchestrator/models.go
```diff
@@ -74,3 +84,103 @@ func AIModelDetails(modelsIndex *modelsindex.ModelsIndex, id string) (AIModelIte
...
+	customModelDescriptor := custommodel.ModelDescriptor{
+		ID:          id,
+		Name:        project.Name,
+		Description: project.Description,
+		Metadata: map[string]string{
+			"source":        "edgeimpulse",
+			"ei-project-id": fmt.Sprintf("%d", projectID),
+			"ei-impulse-id": fmt.Sprintf("%d", impulseID),
+			"ei-model-type": modelType,
+			"ei-engine":     engine,
```
dido18: I would add the last modified at of the project.

internal/api/handlers/ai_models.go
```diff
@@ -16,16 +16,31 @@
...
 	"github.com/arduino/arduino-app-cli/internal/render"
 )
 
+type InstallEIModelRequest struct {
+	ProjectID  int    `json:"project_id" description:"Edge Impulse project ID" example:"123456" required:"true"`
+	ImpulseID  int    `json:"impulse_id" description:"Edge Impulse impulse ID" example:"1" required:"true"`
+	PrjApiKey  string `json:"prj_api_key" description:"Edge Impulse API token" example:"your_edge_impulse_api_token" required:"true"`
+	ModelType  string `json:"model_type" description:"Type of model to build (e.g., 'object-detection')" example:"object-detection" required:"true"`
+	Engine     string `json:"engine" description:"Model engine (e.g., 'tensorflow-lite')" example:"tensorflow-lite" required:"true"`
+	DeviceType string `json:"device_type" description:"Device type for deployment (e.g., 'arduino-uno-q')" example:"arduino-uno-q" required:"true"`
```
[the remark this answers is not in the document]
giulio93: Agreed ,  i left impulseID as required since EI is not clear on how he establish the default, if more than one impulse and the impulseID is not detected i want to return an error in order to avoid confusion

internal/api/handlers/ai_models.go
```diff
@@ -95,3 +105,62 @@ func HandlerDeleteModelByID(dockerClient command.Cli, cfg config.Configuration,
...
+		render.EncodeResponse(w, http.StatusOK, eiModel)
+	}
+}
+
+func (r InstallEIModelRequest) Validate() error {
+	if r.ProjectID <= 0 {
+		return fmt.Errorf("project_id must be a positive integer")
+	}
+	if r.ImpulseID < 0 {
+		return fmt.Errorf("impulse_id must be a non-negative integer")
```
dido18: I think we ned to disambiguate from a missing "impulseId"  (nil) and the "impulseid=0".
If  0 is a valid impulseID, I suggest using pointer fields to detect “missing vs zero”.

Add also unit test for the `Validate` function

internal/api/handlers/ai_models.go
```diff
@@ -95,3 +105,66 @@ func HandlerDeleteModelByID(dockerClient command.Cli, cfg config.Configuration,
...
```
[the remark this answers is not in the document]
giulio93: applied

internal/orchestrator/models.go
```diff
@@ -189,3 +193,120 @@ func checkForModelReferences(ctx context.Context, dockerClient command.Cli, cfg
...
+			"ei-last-modified": project.LastModified.Local().Format("2006-01-02 15:04:05"),
+		},
+		Bricks: buildBrickConfigForEIModel(bricksIndex, project.Category, edgeModelsDir, blobModelsDir),
+	}
+
+	aimodel, err := custommodel.Store(edgeModelsDir, customModelDescriptor, modelRC, "model.eim")
+	if err != nil {
+		if errors.Is(err, syscall.ENOSPC) {
+			return nil, ErrInsufficientStorage
+		}
```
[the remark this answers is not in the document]
giulio93: Unfortunately as you already discover `Content-Type:application/octet-stream`

internal/api/docs/openapi.yaml
```diff
@@ -970,6 +970,47 @@ paths:
       summary: Get AI model details
       tags:
       - AIModels
+  /v1/models/edge-impulse:
```
lucarin91: Maybe it's too late, but I would rename in something like
 ```suggestion
  /v1/models/download/edge-impulse:
```

Comment (giulio93) [reply internal/api/docs/openapi.yaml]: I actually don't, since the get /models return the list of models, having /models/edge-impulse is misleading, since this does not return the list of edge-impulse models but download the models directly

**Review by dido18 — APPROVED** [review]
We can merge this and rework if needed later

File: internal/api/handlers/ai_models.go [hunk 7]
```diff
@@ -95,3 +103,73 @@ func HandlerDeleteModelByID(dockerClient command.Cli, cfg config.Configuration,
...
+				slog.Error("forbidden access to Edge Impulse model", slog.String("error", err.Error()))
+				render.EncodeResponse(w, http.StatusForbidden, models.ErrorResponse{Details: "forbidden access to Edge Impulse model"})
+				return
+			case errors.Is(err, orchestrator.ErrInsufficientStorage):
+				slog.Error("insufficient storage to install Edge Impulse model", slog.String("error", err.Error()))
+				render.EncodeResponse(w, http.StatusInsufficientStorage, models.ErrorResponse{Details: "insufficient storage to install Edge Impulse model"})
+				return
+			default:
+				slog.Error("unable to install Edge Impulse model", slog.String("error", err.Error()))
+				render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "unable to install Edge Impulse model: " + err.Error()})
```
lucarin91: This code is really hard to read. We should reuse the same approach we did in the update api https://github.com/arduino/arduino-app-cli/blob/e91e8aa3bc6e9c95e37befb061ccb935009cd53b/internal/update/errors.go

PR #216, opened 2026-01-28
exported app zip should have a root folder

### LINKED ISSUE #215: exported app zip should have a root folder
### Describe the problem

There is a common convention that a zip with multiple files has a root folder, to make allo people to extract the archives without worring of potencial file override.

adds support for “**smart import**”, accepting the following use cases:
flat.zip -> flat.zip/app.yamlroot.zip -> root.zip/my-app/app.yamldeep_nested.zip -> deep_nested.zip/nested/root/app.yaml. deep_nested.zip/nested/....../root/app.yaml
However, any app folder structure, even when nested, must follow the rules described here : https://github.com/arduino/arduino-app-cli/pull/67/changes

### To reproduce

```
 curl http://localhost:8800/v1/apps/<app-name>/export --output asd.zip

unzip -l asd.zip
```

### Expected behavior

The zip should have a folder with the name of the app; for instance, the name isn't important.

### Arduino App CLI version

0.8.2

### Additional context

_No response_

### Issue checklist

mirkoCrobu:
### Motivation
closes #215 

There is a common convention that a zip with multiple files has a root folder, to allow all people to extract the archives without worrying about potential file override.

### Change description

Added root folder for the app. Root folder name matches app name

#### Smart Import Implementation
Updated the import logic to automatically detect the application root within the ZIP archive (by locating app.yaml). This ensures full compatibility with the new nested export format, flat format(previous one), and deeply nested structures format.

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/archive.go
+++ b/internal/orchestrator/archive.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


lucarin91 (dismissed):

internal/orchestrator/archive.go
```diff
@@ -25,6 +25,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path"
```
MatteoPologruto: Minor, but I think we can avoid using this library and keep using `github.com/arduino/go-paths-helper` where needed for consistency.

internal/orchestrator/archive.go
```diff
@@ -348,3 +395,25 @@ func redactSecrets(bricksindex *bricksindex.BricksIndex, desc *app.AppDescriptor
 		}
 	}
 }
+
+func findZipRoot(r *zip.Reader) (string, error) {
+	for _, f := range r.File {
+		name := filepath.ToSlash(f.Name)
+		if !strings.HasSuffix(name, "app.yaml") && !strings.HasSuffix(name, "app.yml") {
```
Xayton: I think this code is missing a separator in the searched strings.
In other words it would match files like `somethingapp.yaml`, which is not what we want.

MatteoPologruto: We could probably also write a similar check using `paths.New(name).Base() != "app.yaml`.

PR #223, opened 2026-02-03
fix(pkg/board): make sure to close the cmd in case of error

lucarin91:
### Motivation

We need to make sure to always close the cmd process to release all the os resources.

### Reviewer checklist

### DIFF
--- a/pkg/board/remote/adb/adb_nowindows.go
+++ b/pkg/board/remote/adb/adb_nowindows.go
@@ -75,6 +75,7 @@ func adbWriteFile(a *ADBConnection, r io.Reader, pathStr string) error {
 	if err := cmd.Start(); err != nil {
 		return fmt.Errorf("failed to start command for write file %q: %w", pathStr, err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/board/remote/adb/adb_nowindows.go
```diff
@@ -71,9 +71,11 @@ func adbWriteFile(a *ADBConnection, r io.Reader, pathStr string) error {
 		return fmt.Errorf("failed to get stdin pipe for command to write file %q: %w", pathStr, err)
 	}
 	defer stdin.Close()
+
 	if err := cmd.Start(); err != nil {
 		return fmt.Errorf("failed to start command for write file %q: %w", pathStr, err)
 	}
+	defer func() { _ = cmd.Wait() }()
```
MatteoPologruto:
```suggestion
	// Close cmd regardless of errors happening downstream
	defer func() { _ = cmd.Wait() }()
```
Can we add a small comment to explain why we are doing this?

PR #227, opened 2026-02-04
Download libraries used in examples during system init

cmaglie:
### Motivation

Now `system init` downloads and installs libraries used in examples.

### Change description

The libraries are installed in the `internal` section of Arduino CLI's data directory.

### Reviewer checklist


lucarin91 (commented):
I think we should add a callback for getting some progress. Or at least prepare the way to do it. Ideally, we should implement a progress bar in the terminal for reporting the download status.

PR #242, opened 2026-02-11
Resolve gid using the host values.

martacarbone:
### Motivation

Resolve group IDs on the host system before passing them to the container. This ensures that the process running inside the container belongs to the correct group within the host namespace

Screenshot 2026-02-11 at 18 15 41

Screenshot 2026-02-11 at 21 05 40

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/provision.go
+++ b/internal/orchestrator/provision.go
@@ -438,7 +438,7 @@ func lookupGroups(groupNames ...string) []uint32 {
 	for _, name := range groupNames {
 		g, err := user.LookupGroup(name)
 		if err != nil {
-			slog.Warn("Warning: group not found on host; skipping", "group", name)
+			slog.Warn("group not found on host; skipping", "group", name)
 			continue
 		}
 		gid, err := strconv.ParseUint(g.Gid, 10, 32)


internal/orchestrator/provision.go
```diff
@@ -428,6 +429,16 @@ func generateMainComposeFile(
 	return nil
 }
 
+// Resolve a group dynamically, as its GIDs may differ between
+// the host and container environments.
+func lookupGroupId(groupName string) string {
+	g, err := user.LookupGroup(groupName)
+	if err != nil {
+		panic(err)
```
lucarin91: This means that we need the groups named `render` and `gpiod`. Should we include them in the installation process?

lucarin91:
@martacarbone sorry in the second image, why there is a 4444 group?

internal/orchestrator/provision.go
```diff
@@ -349,7 +350,7 @@ func generateMainComposeFile(
 
 	volumes = addLedControl(volumes)
 
-	groups := []string{"dialout", "video", "audio", "render"}
+	groups := []string{"dialout", "video", "audio", "render", lookupGroupId("render")}
```
lucarin91: I think we should only use id for group, so we are sure we are passing the host one

internal/orchestrator/provision.go
```diff
@@ -428,6 +430,28 @@ func generateMainComposeFile(
 	return nil
 }
 
+// Resolve supplementary group IDs on the host dynamically
+// before assigning them to the container, as numeric GIDs
+// could differ between host and container environments.
+func lookupGroups(groupNames []string) []uint32 {
```
lucarin91:
```suggestion
func lookupGroups(groupNames ...string) []uint32 {
```

PR #249, opened 2026-02-14
The platform update should not return error if no packages found

giulio93:
### Motivation
During `system update` the cli return an error saying "0 packages found but one was expected", even if there is not platform update. 
If there is no platform update, we should consider the request successfull.


internal/update/arduino/arduino.go
lucarin91: I would return early here directly if the packages list is empty

PR #256, opened 2026-02-18
fix: move require device check out of provisioning

martacarbone:
### Motivation

Extracted and moved device functions from provision and orchestrator into a new device.go file.
Applied a small fix to extractIndexFromVideoDeviceName and added TestExtractIndexFromVideoDeviceName to verify it.

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -128,6 +128,7 @@ func StartApp(
 		devices, err := peripherals.Detect()
 		if err != nil {
 			yield(StreamMessage{error: err})
+			return
 		}
 
 		err = checkRequiredDevices(bricksIndex, appToStart.Descriptor.Bricks, devices)


internal/orchestrator/app/devices_test.go
```diff
@@ -0,0 +1,145 @@
...
+	sortV4lByIndexDevices(devices)
+	assert.Equal(t, "usb-Generic_GENERAL_-_UVC-video-index0", devices[0])
+	assert.Equal(t, "usb-Generic_GENERAL_-_UVC-video-index1", devices[1])
+	assert.Equal(t, "usb-046d_0825-video-index2", devices[2])
+}
+
+func TestValidateRequiredDevice(t *testing.T) {
+	testCases := []struct {
+		name            string
+		requiredDevices []string
```
dido18: requiredDevices is not used, you can remove it

internal/orchestrator/peripherals/devices_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSortV4LVideoDevices(t *testing.T) {
+
```
dido18: nit
```suggestion
```

internal/orchestrator/app/validator.go
```diff
@@ -54,3 +57,55 @@ func ValidateBricks(a AppDescriptor, index *bricksindex.BricksIndex, modelIndex
...
 }
+
+const (
+	CameraDevice     = "camera"
+	MicrophoneDevice = "microphone"
+	SpeakerDevice    = "speaker"
+)
+
+func ValidateRequiredDevices(bricksIndex *bricksindex.BricksIndex, appBricks []Brick, res *peripherals.AvailableDevices) error {
+	// A brick defines a list of required device classes
```
dido18:
```suggestion
```

internal/orchestrator/app/validator.go
```diff
@@ -54,3 +57,55 @@ func ValidateBricks(a AppDescriptor, index *bricksindex.BricksIndex, modelIndex
...
+	requiredDeviceClasses := make(map[string]bool)
+
+	for _, brick := range appBricks {
+		idxBrick, found := bricksIndex.FindBrickByID(brick.ID)
+		if !found {
+			slog.Warn("Brick not found", slog.String("brick_id", brick.ID))
+			continue
+		}
+
```
dido18:
```suggestion
	// A brick defines a list of required device classes
```

internal/orchestrator/app/validator.go
```diff
@@ -54,3 +57,55 @@ func ValidateBricks(a AppDescriptor, index *bricksindex.BricksIndex, modelIndex
...
+)
+
+func ValidateRequiredDevices(bricksIndex *bricksindex.BricksIndex, appBricks []Brick, res *peripherals.AvailableDevices) error {
+	// A brick defines a list of required device classes
+	requiredDeviceClasses := make(map[string]bool)
+
+	for _, brick := range appBricks {
+		idxBrick, found := bricksIndex.FindBrickByID(brick.ID)
+		if !found {
+			slog.Warn("Brick not found", slog.String("brick_id", brick.ID))
```
dido18:
```suggestion
			slog.Warn("Cannot validate required devices. Brick not found", slog.String("brick_id", brick.ID))
```

internal/orchestrator/app/validator_test.go
```diff
@@ -195,3 +196,168 @@ bricks:
...
+			hasSoundDevice:  false,
+			hasGPUDevice:    false,
+			errMessage:      "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			// build the index
```
dido18: nit
```suggestion
```

internal/orchestrator/peripherals/devices.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+type AvailableDevices struct {
+	DevicePaths    []string
+	HasVideoDevice bool
+	HasSoundDevice bool
+	HasGPUDevice   bool
+}
+
+func GetAvailableDevices() (*AvailableDevices, error) {
```
dido18: naming is hard, but maybe something `Detect()` so that the usage will be `peripheral.Detect()` ?

not a strong opinion on this, only to have cleaner api interface

internal/orchestrator/peripherals/devices.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+type AvailableDevices struct {
+	DevicePaths    []string
+	HasVideoDevice bool
+	HasSoundDevice bool
+	HasGPUDevice   bool
+}
+
+func Detect() (*AvailableDevices, error) {
```
lucarin91: Avoid pointers if not necessary. Pointer usually suggests that the value could be nil, but it doesn't seem to be the case in this function; either you return a valid value or an error.

```suggestion
func Detect() (AvailableDevices, error) {
```

internal/orchestrator/app/validator.go
```diff
@@ -54,3 +57,54 @@ func ValidateBricks(a AppDescriptor, index *bricksindex.BricksIndex, modelIndex
...
 	return allErrors
 }
+
+const (
+	CameraDevice     = "camera"
+	MicrophoneDevice = "microphone"
+	SpeakerDevice    = "speaker"
+)
+
+func ValidateRequiredDevices(bricksIndex *bricksindex.BricksIndex, appBricks []Brick, res *peripherals.AvailableDevices) error {
```
lucarin91: This doesn't seem like a validation function. Validation seems like a static analysis, here you are checking if you have all the required peripehral curretly connected to your system, right? I will call this `CheckRequiredDevices` and I would move it out from the validator.go file and probably outside of the `app` module. I would move into `orchestrator.go`

dido18 (approved):

PR #258, opened 2026-02-19
Updated `licensed` task and fixed license cache

cmaglie:
### Motivation

Updated the license check task so it won't allow `other` during policy check.

### Additional Notes

https://github.com/arduino/tooling-project-assets/issues/683

### Reviewer checklist

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -267,22 +267,20 @@ tasks:
     cmds:
       - |
           MATCHES="$(
-            go tool \
-              github.com/mikefarah/yq/v4 \
-                'select(.license == "other") | filename' \
-                $(
-                  find \
-                    .licenses \
-                    -regex '.+\.yml' \
-                    -type f \
-                    -print0 \
-                  | \
-                    xargs \
-                      --null
-                )
+            find \
+              .licenses \
+              -regex '.+\.yml' \
+              -type f \
+              -print0 \
+            | \
+              xargs \
+                --null \
+                go tool \
+                  github.com/mikefarah/yq/v4 \
+                    'select(.license == "other") | filename'
           )"
           if [[ "$MATCHES" != "" ]]; then
-            echo "$MATCHES"
+            printf "%s" "$MATCHES"
             echo
             echo "The license cache files above have the license set to 'other'. Please manually define the license type and retry."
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -265,6 +265,29 @@ tasks:
...
+                    -regex '.+\.yml' \
+                    -type f \
+                    -print0 \
+                  | \
+                    xargs \
+                      --null
+                )
+          )"
+          if [[ "$MATCHES" != "" ]]; then
+            echo "$MATCHES"
```
per1234:
```suggestion
      - |
        MATCHES="$(
          find \
            .licenses \
            -regex '.+\.yml' \
            -type f \
            -print0 \
          | \
            xargs \
              --null \
              go tool \
                github.com/mikefarah/yq/v4 \
                  'select(.license == "other") | filename'
        )"
        if [[ "$MATCHES" != "" ]]; then
          printf "%s" "$MATCHES"
```

## Correct usage of `xargs`

Previously `xargs` had no effect and the full list of files were always passed as an argument to `yq` (my mistake). This could cause the command to fail with an "`Argument list too long`" error.

After this change, `xargs` makes multiple `yq` invocations with a split argument list as needed to avoid exceeding the system limit.

## Use `printf` instead of `echo`

I found that, after the above change, the output from the task would be corrupted (seems like it was not doing a carriage return). For example:

  ```text
   indows.dep.ymlo/github.com/arduino/arduino-app-cli/pkg/board.dep.yml
                                                                       .licenses/arduino-app-cli/go/github.com/arduino/arduino-app-cli/pkg/x/ports.dep.yml
                                                                                                                                                          .licenses/arduino-app-cli/go/golang.org/x/sys/windows/registry.dep.yml
                                                                                                                                                                                                                                .licenses/arduino-app-cli/go/golang.org/x/sys/w
                 The license cache files above have the license set to 'other'. Please manually define the license type and retry.
                                                                                                                                  See: https://github.com/arduino/tooling-project-assets/blob/main/workflow-templates/check-go-dependencies-task.md#unrecognized-licenses
   ```

I'm not sure exactly what the problem was, but I found that when I use `printf "%s"`, the output is as expected. The fault still occurred if I did not provide `printf` with a format specifier.

per1234 (changes requested):

PR #269, opened 2026-02-24
feat: extract platform dependent stuff

lucarin91:
### Motivation

Move all board-specific configuration to a new platform module. This will improve the portability of the software.

### Reviewer checklist

### DIFF
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -34,6 +34,7 @@ func GetPlatform() Platform {
 	codeName := getCodeName()
 	switch codeName {
 	case "Imola":
+		slog.Debug("detected platform", "codeName", codeName)
 		return Platform{
 			codeName:   codeName,
 			FQBN:       "arduino:zephyr:unoq",


internal/platform/platform.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type GpioPin struct {
+	Chip   string
+	Number int
+}
+
+type Platform struct {
+	BoardName    string
+	FQBN         string
```
dido18: Is this `BoardFQBN`?

internal/platform/platform.go
```diff
@@ -0,0 +1,85 @@
...
+		slog.Warn("not supported platform", "boardName", boardName)
+		return Platform{
+			BoardName: boardName,
+		}
+	}
+}
+
+func (p Platform) GetMicro() micro.Micro {
+	return micro.New(micro.GpioPin(p.Micro.ResetPin))
+}
```
dido18: is there any reason why to not have the `micro.Micro` struct in the `Platform` instead of building it in the `GetMicro`  call ?

internal/platform/platform.go
dido18: Maybe it is clearer to have `BoardFQBN` . 
The usage will be `platform.BoardFQBN` that is the FQBN of the board (not of the platform)

internal/platform/platform.go
lucarin91: because micro has some linux dependend code and I didn't want to pollute the platform with that

dido18 (approved):

PR #254, opened 2026-02-17
feat: compile with new wait_for_app mode

lucarin91:
### Motivation

Implement the new wait_for_app menu https://github.com/arduino/ArduinoCore-zephyr/pull/369

### Additional Notes

- ~wait for https://github.com/arduino/ArduinoCore-zephyr/compare/main...arduino:ArduinoCore-zephyr:wait_for_app_mode~ 

### Reviewer checklist

- [ ] PR addresses a single concern.
- [ ] PR title and description are properly filled.
- [ ] Changes will be merged in `main`.
- [ ] Changes are covered by tests.
- [ ] Logging is meaningful in case of troubleshooting.

PR #270, opened 2026-02-25
chore: block platform update to 0.53.1

lucarin91:
### Motivation

The Zephyr platform will dismiss flash to RAM in the next release, so we will need to prevent update till we implement an alternative solution, like this https://github.com/arduino/arduino-app-cli/pull/254
 
### Reviewer checklist


lucarin91:
> @lucarin91 Can we push to have the #254 merged before the next platform update? Or is there a reason why this is not feasible right now?

That one isn't currently supported, and the two features will not be supported at the same time, so we probably need to be smarter on our side. Probably this isn't enough either. Because after we unblock the constraint, we should still support both changes till the user really updates the platform.

PR #283, opened 2026-03-04
feat: support local brick

dido18:
### Motivation
Support a local bricks inside the app folder.

### Change description
Added a brick `Manager` that has a list of Brick sources. A brick source implements the following api
- `GetByID(id string) (*Brick, bool)`
- `ListBricks() ([]Brick, bool)`
- `GetReadme(id string) (string, error)`
- `GetComposePath(id string) (*paths.Path, bool)`
- `GetApiDocPath(id string) (*paths.Path, error)`
- `GetExamplesPath(id string) (paths.PathList, error)`

Currenlty, there are two brick sources:
- `builtin` that loads the bricks index from `brick-list.yaml` file
- `applocal` that loads the bricks index from an app

The applocal index is added to the manager only when an app is present.

### Reviewer checklist

### DIFF
--- a/internal/api/handlers/app_ports.go
+++ b/internal/api/handlers/app_ports.go
@@ -53,7 +53,12 @@ func HandleAppPorts(
 			render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "unable to find the app"})
 			return
 		}
-		brickInfoMap, err := GetBrickPortInfoByID(app.Descriptor.Bricks, bricksIndex)
+		localIndex, err := app.LoadAppBricksIndex()
+		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/orchestrator/brickslocalindex/loader.go
```diff
@@ -0,0 +1,90 @@
...
+		if file.Join("brick_config.yaml").NotExist() {
+			// let's continue scanning, the model can be in a subfolder
+			return true
+		}
+		return false
+	}, paths.FilterDirectories())
+	if err != nil {
+		return nil, err
+	}
+	bricks := []AppLocalBrick{}
```
mirkoCrobu:
```suggestion
bricks := make([]AppLocalBrick, 0, len(pathsList))
```

PR #288, opened 2026-03-09
Edge Impulse custom AI model

mirkoCrobu:
### Motivation

- [x] Resolve the TODO isValid` for custom model ([code](https://github.com/arduino/arduino-app-cli/blob/043b86b47ca9596beb02816bbfe9936a0ceaea9d/internal/orchestrator/modelsindex/custommodel/parser.go#L38))

### Reviewer checklist

### DIFF
--- a/internal/e2e/daemon/ai_models_test.go
+++ b/internal/e2e/daemon/ai_models_test.go
@@ -104,7 +104,8 @@ func TestAIModelDetails(t *testing.T) {
 	t.Run("should return full details for a valid custom model ID", func(t *testing.T) {
 		_, err := custommodel.Store(customModelDir.Join("my-model"), custommodel.ModelDescriptor{
 			ID:          "custom-classification-model-eim",
-			Name:        "this the name of the model",
+			Name:        "this is the name of the model",
+			Runner:      "brick",
 			Description: "this is the description of the model",
 			Bricks: []custommodel.BrickConfig{
 				{ID: "arduino:audio_classification"},
@@ -120,7 +121,7 @@ func TestAIModelDetails(t *testing.T) {
 		got := response.JSON200
 		require.Equal(t, &client.AIModelItem{
 			Id:          new("custom-classification-model-eim"),
-			Name:        new("this the name of the model"),
+			Name:        new("this is the name of the model"),
 			IsBuiltin:   new(false),
 			Runner:      new(""),
 			Description: new("this is the description of the model"),
@@ -218,6 +219,7 @@ func TestAIModelDelete(t *testing.T) {
 
 		_, err := custommodel.Store(customModelDir.Join("my-custom-model"), custommodel.ModelDescriptor{
 			ID:     modelId,
+			Name:   "this the name of the model",
 			Runner: "brick",
 			Bricks: []custommodel.BrickConfig{
 				{ID: "arduino:audio_classification"},
--- a/internal/orchestrator/modelsindex/custommodel/custommodel.go
+++ b/internal/orchestrator/modelsindex/custommodel/custommodel.go
@@ -98,8 +98,7 @@ func (a *AiModel) GetDescriptorPath() *paths.Path {
 }
 
 func (a *AiModel) writeDescriptorFile() error {
-	err := a.ModelDescriptor.Validate()
-	if err != nil {
+	if err := a.ModelDescriptor.Validate(); err != nil {
 		return fmt.Errorf("invalid model descriptor: %w", err)
 	}
 
@@ -115,7 +114,7 @@ func (a *AiModel) writeDescriptorFile() error {
 
 	if err := fatomic.WriteFile(descriptorPath.String(), out, os.FileMode(0644)); err != nil {
 		_ = descriptorPath.Remove()
-		return fmt.Errorf("failed to write model descriptorfile: %w", err)
+		return fmt.Errorf("failed to write model descriptor file: %w", err)
 	}
 	return nil
 }
--- a/internal/orchestrator/modelsindex/custommodel/custommodel.go
+++ b/internal/orchestrator/modelsindex/custommodel/custommodel.go
@@ -98,10 +98,11 @@ func (a *AiModel) GetDescriptorPath() *paths.Path {
 }
 
 func (a *AiModel) writeDescriptorFile() error {
-	if !a.ModelDescriptor.IsValid() {
-		// TODO: provide more details about the invalidity
-		return errors.New("invalid model descriptor")
+	err := a.ModelDescriptor.Validate()
+	if err != nil {
+		return fmt.Errorf("invalid model descriptor: %w", err)
 	}
+
 	descriptorPath := a.GetDescriptorPath()
 	if descriptorPath == nil {
 		return errors.New("model descriptor file path is not set")
--- a/internal/orchestrator/modelsindex/custommodel/parser.go
+++ b/internal/orchestrator/modelsindex/custommodel/parser.go
@@ -61,6 +61,7 @@ func (a *ModelDescriptor) Validate() error {
 	return err
 }
 
+// validateEdgeImpulseMetadata checks that all metadata fields required by App Lab are present and valid.
 func validateEdgeImpulseMetadata(metadata map[string]string) error {
 	requiredFields := []string{
 		"ei-project-id",
--- a/internal/orchestrator/modelsindex/custommodel/parser.go
+++ b/internal/orchestrator/modelsindex/custommodel/parser.go
@@ -7,9 +7,12 @@ package custommodel
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/arduino/go-paths-helper"
 	"github.com/goccy/go-yaml"
+
+	"github.com/arduino/arduino-app-cli/internal/orchestrator/bricksindex"
 )
 
 type ModelDescriptor struct {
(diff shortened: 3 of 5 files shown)


internal/e2e/daemon/ai_models_test.go
```diff
@@ -174,6 +174,7 @@ func TestAIModelDetails(t *testing.T) {
 }
 
 func TestAIModelDelete(t *testing.T) {
+
```
dido18:
```suggestion
```

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -52,10 +55,64 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
...
+	return nil
+}
+
+func validateEdgeImpulseMetadata(metadata map[string]string) error {
+	requiredFields := []string{
+		"ei-project-id",
+		"ei-impulse-id",
+		"ei-impulse-name",
+		"ei-deployment-version",
+	}
```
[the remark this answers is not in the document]
mirkoCrobu: for example I see from App-lab that they read ei-deployment-version to check if an installed model is outdated:

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -52,10 +55,64 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
...
+		}
+	}
+
+	if metadata["ei-model-type"] != "float32" {
+		return fmt.Errorf("invalid Edge Impulse metadata: unsupported model type")
+	}
+
+	if metadata["ei-engine"] != "tflite" {
+		return fmt.Errorf("invalid Edge Impulse metadata: unsupported engine")
+	}
```
dido18, line 84: Are these checks true?  Are you sure only float32 is supported?

mirkoCrobu: yes for EI. maybe different for others custom models.

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -52,10 +55,64 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
...
+	if !ok {
+		return nil // source is optional
+	}
+
+	switch source {
+	case "edgeimpulse":
+		return validateEdgeImpulseMetadata(a.Metadata)
+	default:
+		return fmt.Errorf("invalid model descriptor: unsupported source '%s'", source)
+	}
```
lucarin91: I would return all the validation error not just the first one
```suggestion
    var err error
	if a.ID == "" {
		err = errors.Join(err, fmt.Errorf("invalid model descriptor: id is empty"))
	}
	if a.Name == "" {
		err = errors.Join(err, fmt.Errorf("invalid model descriptor: name is empty"))
	}
	source, ok := a.Metadata["source"]
	if ok {
	    switch source {
	    case "edgeimpulse":
		    err = errors.Join(err, validateEdgeImpulseMetadata(a.Metadata))
	    default:
		    err = errors.Join(fmt.Errorf("invalid model descriptor: unsupported source '%s'", source))
	    }
	}
	return err
```

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -52,10 +55,64 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
...
+			return fmt.Errorf("invalid model descriptor: brick with ID '%s' not found", brickConfig.ID)
+		}
+
+		for _, variable := range brick.Variables {
+			if strings.HasPrefix(variable.Name, "EI_") && strings.HasSuffix(variable.Name, "_MODEL") {
+				if val, ok := brickConfig.ModelConfiguration[variable.Name]; !ok || val == "" {
+					return fmt.Errorf("invalid model descriptor: missing model configuration for variable '%s' in brick '%s'", variable.Name, brickConfig.ID)
+				}
+			}
+		}
```
lucarin91:
```suggestion
        var err error
		for _, variable := range brick.Variables {
			if strings.HasPrefix(variable.Name, "EI_") && strings.HasSuffix(variable.Name, "_MODEL") {
				if val, ok := brickConfig.ModelConfiguration[variable.Name]; !ok || val == "" {
					err = errors.Join(err, fmt.Errorf("invalid model descriptor: missing model configuration for variable '%s' in brick '%s'", variable.Name, brickConfig.ID))
				}
			}
		}
```

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -52,10 +55,64 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
...
+	}
+	return nil
```
lucarin91:
```suggestion
	return err
```

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -52,10 +55,64 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
...
+			return fmt.Errorf("invalid Edge Impulse metadata: missing required field '%s'", field)
+		}
+	}
+
+	if metadata["ei-model-type"] != "float32" {
+		return fmt.Errorf("invalid Edge Impulse metadata: unsupported model type")
+	}
+
+	if metadata["ei-engine"] != "tflite" {
+		return fmt.Errorf("invalid Edge Impulse metadata: unsupported engine")
```
lucarin91:
```suggestion
   var err err
	for _, field := range requiredFields {
		if val, ok := metadata[field]; !ok || val == "" {
			err = errors.Join(err, fmt.Errorf("invalid Edge Impulse metadata: missing required field '%s'", field))
		}
	}

	if metadata["ei-model-type"] != "float32" {
		err = errors.Join(err, fmt.Errorf("invalid Edge Impulse metadata: unsupported model type"))
	}

	if metadata["ei-engine"] != "tflite" {
		err = errors.Join(err, fmt.Errorf("invalid Edge Impulse metadata: unsupported engine"))
```

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -52,10 +55,64 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
...
+
+	switch source {
+	case "edgeimpulse":
+		return validateEdgeImpulseMetadata(a.Metadata)
+	default:
+		return fmt.Errorf("invalid model descriptor: unsupported source '%s'", source)
+	}
+}
+
+func (a *ModelDescriptor) CheckEdgeImpulseBricks(bricksIndex *bricksindex.BricksIndex) error {
```
lucarin91: This function seems not to be used, do we need to call it somewhere?

lucarin91 (commented):
I would try to return as many errors as possible instead of stopping at the first one

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -40,10 +41,45 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
 	return descriptor, nil
 }
 
-func (a *ModelDescriptor) IsValid() bool {
-	/*  TODO: check
-	1) brick list are present into the brick-list
-	2) metadata are coherent with the source
-	*/
-	return true
+func (a *ModelDescriptor) Validate() error {
```
dido18, line 44: do we have unit test for the Validate function ?

internal/orchestrator/check.go
```diff
@@ -34,36 +34,47 @@ func checkBricks(a app.AppDescriptor, index *bricksindex.BricksIndex, modelIndex
...
 			}
 		}
 
-		// Check that all required brick variables are provided by app
+		// Check that all required brick variables are provided by app or by the model
 		for _, indexBrickVariable := range indexBrick.Variables {
-			if indexBrickVariable.IsRequired() {
-				if _, exist := appBrick.Variables[indexBrickVariable.Name]; !exist {
-					allErrors = errors.Join(allErrors, fmt.Errorf("variable %q is required by brick %q", indexBrickVariable.Name, indexBrick.ID))
+			if !indexBrickVariable.IsRequired() {
```
dido18: Is there any reason to have these changes in this PR, if not i would move this changes into another separate PR and leave here only the `validate` function

mirkoCrobu: no. Removed. we don't need it since we don't have real usecases for this check.

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -40,10 +41,45 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
 	return descriptor, nil
 }
 
-func (a *ModelDescriptor) IsValid() bool {
-	/*  TODO: check
-	1) brick list are present into the brick-list
```
mirkoCrobu, line 45: The TODO about checking that brick IDs exist in the brick index is covered by `buildBrickConfigForEIModel`(that, together with Validate(), is part of the same installEImodel flow), which validates brick existence before the descriptor is even built. 
This means the validation step is split into two different functions. This could be a problem if Store() that contains just Validate() will be called in a different flow without `buildBrickConfigForEIModel`. 

Should we open a separate issue to evaluate a refactoring?

Xayton (approved):

internal/orchestrator/modelsindex/custommodel/custommodel.go
```diff
@@ -98,10 +98,11 @@ func (a *AiModel) GetDescriptorPath() *paths.Path {
 }
 
 func (a *AiModel) writeDescriptorFile() error {
-	if !a.ModelDescriptor.IsValid() {
-		// TODO: provide more details about the invalidity
-		return errors.New("invalid model descriptor")
+	err := a.ModelDescriptor.Validate()
+	if err != nil {
```
lucarin91: nit
```suggestion
	if err := a.ModelDescriptor.Validate(); err != nil {
```

internal/orchestrator/modelsindex/custommodel/parser.go
```diff
@@ -40,10 +41,45 @@ func ParseModelDescriptorFile(file *paths.Path) (ModelDescriptor, error) {
...
+		case "edgeimpulse":
+			err = errors.Join(err, validateEdgeImpulseMetadata(a.Metadata))
+		default:
+			err = errors.Join(err, fmt.Errorf("invalid model descriptor: unsupported source '%s'", source))
+		}
+	}
+	return err
+}
+
+func validateEdgeImpulseMetadata(metadata map[string]string) error {
```
lucarin91, line 65: In general, I don't think we should validate metadata. Metadata by design should be a freeform map, so we could have whatever we want there. AFAIK, we don't require any special metadata, and we never read that either.

lucarin91 (changes requested):

internal/orchestrator/modelsindex/custommodel/parser.go
Xayton: But I think it was said that, if some metadata is missing, then we cannot use the AI model.
So I think it helps to know that something is missing, as soon as possible.

internal/orchestrator/modelsindex/custommodel/custommodel.go
mirkoCrobu: done

internal/orchestrator/modelsindex/custommodel/parser.go
mirkoCrobu: They are mandatory for App Lab. I added a comment

dido18 (approved):

lucarin91 (approved):

PR #321, opened 2026-03-30
feat: add support for singleton brick services

MatteoPologruto:
### Motivation
A brick service is a singleton service that the `arduino-app-cli` must instantiate only once and shared among different bricks. A brick that needs a service, can define it in the `brick_config.yaml` this way:
```
requires_services: ["arduino:genie"]
```
During the provisioning process of an app, the `arduino-app-cli` should collect all the shared services needed to run the bricks and instantiate them all once.

### Change description
This PR introduces two main changes:
- a `servicesIndex` is created following the same principles of the `bricksIndex` and `modelsIndex`
- the services are collected and instantiated during the provisioning of the app

### Reviewer checklist


lucarin91 (changes requested):
LGTM, just some small issues

lucarin91 (approved):

dido18 (changes requested):
We need to get the service even in the logs endpoint to retrieve the compose services' logs.

Could you double-check it?

dido18 (approved):
Remember to open a PR to check other usage of docker-compose services

PR #322, opened 2026-03-30
feat: add support for the new platform 0.55

lucarin91:
### Motivation

The new 0.55 platform will auto install the `Arduino_RouterBridge` inside the `arduino:zephyr` with the new `libraryDependencies` feature and so we don't want to pin that library in the sketch profle. 

### Change description

Update the arduino-cli to support `libraryDependencies` and add the migration job, which remove the `Arduino_RouterBridge` library.

### Additional Notes

- wait for https://github.com/arduino/arduino-cli/pull/3142
- wait for https://github.com/arduino/arduino-cli/issues/3155

### Reviewer checklist


internal/orchestrator/orchestrator.go
```diff
@@ -175,7 +176,18 @@ func StartApp(
...
 				return
 			}
-			if err := compileUploadSketch(ctx, platform, &appToStart, sketchCallbackWriter); err != nil {
+
+			if ok, err := migration_to_platform_gt_0_54_1(ctx, platform, appToStart); err != nil {
+				if !yield(StreamMessage{data: "Failed to apply app migration for platform arduino:zephyr >0.54.1. Error: " + err.Error()}) {
+					return
+				}
+			} else if ok {
+				if !yield(StreamMessage{data: "Applied app migration for platform arduino:zephyr >0.54.1. Arduino_RouterBridge is now part of the platform and shouldn't explicit specify"}) {
```
cmaglie:
```suggestion
				if !yield(StreamMessage{data: "Applied app migration for platform arduino:zephyr >0.54.1. Arduino_RouterBridge is now part of the platform and shouldn't be explicitly specified"}) {
```

internal/orchestrator/helpers.go
```diff
@@ -215,3 +217,39 @@ func setStatusLeds(platform platform.Platform, trigger LedTrigger) error {
...
+			EncodedValue: dataDir,
+			ValueFormat:  "cli",
+		})
+		if err != nil {
+			return fmt.Errorf("could not set data directory: %w", err)
+		}
+	}
+
+	// Set the additional urls via ARDUINO_BOARD_MANAGER_ADDITIONAL_URLS env var
+	if urls, err := os.LookupEnv("ARDUINO_BOARD_MANAGER_ADDITIONAL_URLS"); err {
```
mirkoCrobu:
```suggestion
if urls, ok := os.LookupEnv("ARDUINO_BOARD_MANAGER_ADDITIONAL_URLS"); ok {
```

internal/orchestrator/orchestrator.go
```diff
@@ -1181,6 +1196,73 @@ func compileUploadSketch(
 	}, stream)
 }
 
+// migration_to_platform_gt_0_54_1 removes the Arduino_RouterBridge library from the sketch profile to allow automatic update of the library.
+// This is needed by the platform 0.55 will need a new Arduino_RouterBridge library to allow Serial output redirection to Monitor.
+func migration_to_platform_gt_0_54_1(ctx context.Context, platform platform.Platform, app app.ArduinoApp) (bool, error) {
```
mirkoCrobu: Could we use the camelCase convention here? and mayberemoving the version from the func name (we could keep the vers only in the comment)
```suggestion
func migrateRemoveRouterBridgeIfNeeded(ctx context.Context, platform platform.Platform, app app.ArduinoApp) (bool, error) {
```

internal/orchestrator/orchestrator.go
```diff
@@ -175,7 +176,18 @@ func StartApp(
 			if !yield(StreamMessage{progress: &Progress{Name: "sketch compiling and uploading", Progress: 0.0}}) {
 				return
 			}
-			if err := compileUploadSketch(ctx, platform, &appToStart, sketchCallbackWriter); err != nil {
+
+			if ok, err := migration_to_platform_gt_0_54_1(ctx, platform, appToStart); err != nil {
+				if !yield(StreamMessage{data: "Failed to apply app migration for platform arduino:zephyr >0.54.1. Error: " + err.Error()}) {
+					return
```
dido18, line 182: Could we suggest a fix  message here ?

internal/orchestrator/orchestrator.go
```diff
@@ -1181,6 +1196,88 @@ func compileUploadSketch(
...
+	if err != nil {
+		return false, err
+	}
+
+	sketch := sketchResp.GetSketch()
+	platforms := sketch.GetDefaultProfile().GetPlatforms()
+	if slices.ContainsFunc(platforms, func(p *rpc.ProfilePlatformReference) bool {
+		return p.GetVersion() != ""
+	}) {
+		slog.Debug("sketch profile has fixed platform version, skipping migration")
```
dido18: could you clarify the message of this log ? What is a  "fixed" version of the platform defined in the profile.
What happens if the platform is <0.54.1 ?

```sketch.yaml
platform: 0.50.1
```

internal/orchestrator/orchestrator.go
lucarin91: it should be in the compilation error

mirkoCrobu (approved):

PR #347, opened 2026-04-09
fix: return brick info best effor

lucarin91:
### Reviewer checklist


internal/orchestrator/bricks/bricks.go
```diff
@@ -180,33 +180,19 @@ func (s *Service) BricksDetails(id string, idProvider *app.IDProvider,
...
-	if err != nil {
-		return BrickDetailsResult{}, fmt.Errorf("cannot open docs for brick %s: %w", id, err)
-	}
-
-	apiDocsPath, found := brick.GetApiDocPath()
-	if !found {
-		return BrickDetailsResult{}, fmt.Errorf("cannot open api-docs for brick %s", id)
-	}
+	readme, _ := brick.GetReadmeFile()
+	apiDocsPath, _ := brick.GetApiDocPath()
```
mirkoCrobu: apiDocsPath  can be nil. Don't we need to check it before to use it in line 206(`ApiDocsPath:  apiDocsPath.String()`,)?

mirkoCrobu:
What about having at least a "slog.Warn" instead of completely ignoring the errors?

dido18 (approved):

PR #354, opened 2026-04-13
Add leds control functions.

### LINKED ISSUE #262: LED2 system triggers not restored after app stop
## Description

When an app is stopped, `setStatusLeds(LedTriggerDefault)` writes the `"default"` trigger to the LED2 sysfs trigger files (`blue:bt`, `green:wlan`, `red:panic`). However, `"default"` in the Linux kernel LED subsystem corresponds to the `ledtrig-default-on` driver, which simply keeps the LED at its current brightness. It does **not** restore the original hardware triggers (`bluetooth-power`, `phy0tx`, `none`).

As a result, after an app stops:
- **LED2 blue** (`blue:bt`) no longer flashes under `bluetoothd` control
- **LED2 green** (`green:wlan`) no longer blinks on WiFi TX activity
- **LED2 red** (`red:panic`) gets the `default-on` trigger instead of `none`

The LEDs remain in whatever brightness state the app left them in, and the system services (bluetoothd, phy0tx) lose control permanently until reboot.

## Root cause

In [`internal/orchestrator/helpers.go`](https://github.com/arduino/arduino-app-cli/blob/main/internal/orchestrator/helpers.go):

```go
const (
    LedTriggerNone    LedTrigger = "none"
    LedTriggerDefault LedTrigger = "default"  // ← this is the bug
)
```

`setStatusLeds(LedTriggerDefault)` writes `"default"` to each trigger file. The kernel interprets this as the `default-on` trigger, not as "restore to the original/boot trigger".

Additionally, `setStatusLeds(LedTriggerDefault)` is called **before** `docker compose stop`, meaning the container is still running and the app may still be writing to brightness files when the trigger changes.

## Current flow (buggy)

```
START:
  1. setStatusLeds("none")         → disables system triggers :white_check_mark:
  2. addLedControl()               → bind-mounts LED sysfs into container
  3. docker compose up             → app starts writing brightness

STOP:
  1. setStatusLeds("default")      → writes wrong trigger BEFORE container stops :x:
  2. docker compose stop           → SIGTERM to container, app may or may not cleanup
```

## Proposed fix

Replace the fire-and-forget `setStatusLeds` approach with a save/restore pattern, and move the restore to **after** the container has stopped:

```go
type savedLedState struct {
    path    *paths.Path
    trigger string
}

var savedStatusLedTriggers []savedLedState

// saveAndDisableStatusLeds reads the current active trigger for each LED2 channel,
// saves it, and writes "none" to hand control to the app.
func saveAndDisableStatusLeds() error {
    savedStatusLedTriggers = nil
    for _, ledPath := range statusLeds {
        if !ledPath.Exist() {
            return fmt.Errorf("LED path %s does not exist", ledPath)
        }
        content, err := ledPath.ReadFile()
        if err != nil {
            return fmt.Errorf("failed to read %s: %w", ledPath, err)
        }
        trigger := parseActiveTrigger(string(content))
        savedStatusLedTriggers = append(savedStatusLedTriggers, savedLedState{
            path:    ledPath,
            trigger: trigger,
        })
        if err := ledPath.WriteFile([]byte("none")); err != nil {
            return fmt.Errorf("failed to disable trigger for %s: %w", ledPath, err)
        }
    }
    return nil
}

// restoreStatusLeds restores the previously saved triggers and resets brightness to 0.
func restoreStatusLeds() error {
    for _, saved := range savedStatusLedTriggers {
        if err := saved.path.WriteFile([]byte(saved.trigger)); err != nil {
            slog.Warn("failed to restore LED trigger",
                slog.String("path", saved.path.String()),
                slog.String("trigger", saved.trigger),
                slog.String("error", err.Error()))
        }
    }
    // Reset brightness so the trigger driver starts from a clean state
    for _, bp := range statusLedsBrightness {
        if bp.Exist() {
            _ = bp.WriteFile([]byte("0"))
        }
    }
    savedStatusLedTriggers = nil
    return nil
}

// parseActiveTrigger extracts the active trigger from the sysfs trigger file content.
// The active trigger is enclosed in square brackets, e.g. "[bluetooth-power]".
func parseActiveTrigger(content string) string {
    for _, part := range strings.Fields(content) {
        if strings.HasPrefix(part, "[") && strings.HasSuffix(part, "]") {
            return part[1 : len(part)-1]
        }
    }
    return "none"
}
```

In `orchestrator.go`, the stop sequence should be:

```go
// In stopAppWithCmd:
// 1. Do NOT call setStatusLeds here
// 2. Stop MCU (micro.Disable)
// 3. Stop container (docker compose stop — Python cleans up brightness via SIGTERM handler)
// 4. AFTER container stops: restore system triggers
if err := restoreStatusLeds(); err != nil {
    slog.Debug("unable to restore status leds", slog.String("error", err.Error()))
}
```

## Why this is the correct layer

`arduino-app-cli` runs on the **host** and is the component that:
- Knows when apps start and stop
- Disables the system triggers before the container starts
- Has access to sysfs **before and after** the container lifecycle
- Is the single source of truth for system LED management

The Python framework inside the container should only manage **brightness** (what colors/values the app sets), not system triggers — it doesn't know what the original triggers were, and it runs inside a container that may be killed before cleanup completes.

## Expected LED2 triggers on UNO Q at boot

| sysfs path | Boot trigger |
|---|---|
| `/sys/class/leds/blue:bt/trigger` | `bluetooth-power` |
| `/sys/class/leds/green:wlan/trigger` | `phy0tx` |
| `/sys/class/leds/red:panic/trigger` | `none` |

### To reproduce

- start `color-your-leds` example
- set led 1 and 2 ON
- stop the example 

### Expected behavior

led 2 should restore green blinking for wi-fi and blue for bluetooth after app stop.

### Arduino App CLI version

0.8.3-rc.8

### Additional context

_No response_

### Issue checklist

martacarbone:
### Motivation

System leds are not properly reset after usage

Steps to reproduce:
- open and run the "Color your Leds" app
- stop the app
The  two system leds retain its value. The expected behavior is that system leds go back to their initial state (bt, wifi, panic)

After the fix:
- open and run the "Color your Leds" app
- stop the app
- the leds are reset to their default value
- run a test switch on and off the bluetooth light, and a ping will make the wifi led to blink:

```
# bluetooth on
sudo bluetoothctl power on
sleep 1
# bluetooth off
sudo bluetoothctl power off
sleep 1
# wifi should blink
sudo ping -c7 -i 0.5 8.8.8.8
sleep 1
# bluetooth on
sudo bluetoothctl power on

```

### Change description

The change is a brightness and trigger set to default values.

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/helpers.go
+++ b/internal/orchestrator/helpers.go
@@ -203,48 +203,39 @@ const (
 	ledBrightnessTarget ledTarget = "brightness"
 )
 
-type ledTrigger string
-
 const (
-	ledTriggerNone    ledTrigger = "none"
-	ledTriggerDefault ledTrigger = "default"
+	noneTrigger    string = "none"
+	defaultTrigger string = "default"
 )
 
-func setLeds(leds paths.PathList, target ledTarget, value string) error {
-	for _, ledPath := range leds {
-		ledPath = ledPath.Join(string(target))
-		if !ledPath.Exist() {
-			return fmt.Errorf("LED path %s does not exist", ledPath)
-		}
-		if err := ledPath.WriteFile([]byte(value)); err != nil {
-			return fmt.Errorf("failed to set LED %s to %s: %w", ledPath, target, err)
-		}
+func setLeds(ledPath *paths.Path, target ledTarget, value string) error {
+	ledPath = ledPath.Join(string(target))
+	if !ledPath.Exist() {
+		return fmt.Errorf("LED path %s does not exist", ledPath)
+	}
+	if err := ledPath.WriteFile([]byte(value)); err != nil {
+		return fmt.Errorf("failed to set LED %s to %s: %w", ledPath, target, err)
 	}
 	return nil
 }
 
 func setLedsToUserControlledMode(platform platform.Platform) error {
-	if err := setLeds(platform.Linux.StatusLeds, ledTriggerTarget, string(ledTriggerNone)); err != nil {
-		return err
-	}
-	if err := setLeds(platform.Linux.UserLeds, ledTriggerTarget, string(ledTriggerNone)); err != nil {
-		return err
+	for _, led := range slices.Concat(platform.Linux.StatusLeds, platform.Linux.UserLeds) {
+		if err := setLeds(led, ledTriggerTarget, noneTrigger); err != nil {
+			return err
+		}
 	}
 	return nil
 }
 
 func restoreLedsState(platform platform.Platform) error {
-	if err := setLeds(platform.Linux.StatusLeds, ledBrightnessTarget, string("0")); err != nil {
-		return err
-	}
-	if err := setLeds(platform.Linux.UserLeds, ledBrightnessTarget, string("0")); err != nil {
-		return err
-	}
-	if err := setLeds(platform.Linux.StatusLeds, ledTriggerTarget, string(ledTriggerDefault)); err != nil {
-		return err
-	}
-	if err := setLeds(platform.Linux.UserLeds, ledTriggerTarget, string(ledTriggerDefault)); err != nil {
-		return err
+	for _, led := range slices.Concat(platform.Linux.StatusLeds, platform.Linux.UserLeds) {
+		if err := setLeds(led, ledBrightnessTarget, string("0")); err != nil {
+			return err
+		}
+		if err := setLeds(led, ledTriggerTarget, defaultTrigger); err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/internal/orchestrator/helpers.go
+++ b/internal/orchestrator/helpers.go
@@ -196,22 +196,55 @@ func GetCustomErrorFomDockerEvent(message string) error {
 	return nil
 }
 
-type LedTrigger string
+type ledTarget string
 
 const (
-	LedTriggerNone    LedTrigger = "none"
-	LedTriggerDefault LedTrigger = "default"
+	ledTriggerTarget    ledTarget = "trigger"
+	ledBrightnessTarget ledTarget = "brightness"
 )
 
-func setStatusLeds(platform platform.Platform, trigger LedTrigger) error {
-	for _, ledPath := range platform.Linux.StatusLeds {
-		ledPath = ledPath.Join("trigger")
+type ledTrigger string
+
+const (
+	ledTriggerNone    ledTrigger = "none"
+	ledTriggerDefault ledTrigger = "default"
+)
+
+func setLeds(leds paths.PathList, target ledTarget, value string) error {
+	for _, ledPath := range leds {
+		ledPath = ledPath.Join(string(target))
 		if !ledPath.Exist() {
 			return fmt.Errorf("LED path %s does not exist", ledPath)
 		}
-		if err := ledPath.WriteFile([]byte(trigger)); err != nil {
-			return fmt.Errorf("failed to set LED %s to %s: %w", ledPath, trigger, err)
+		if err := ledPath.WriteFile([]byte(value)); err != nil {
+			return fmt.Errorf("failed to set LED %s to %s: %w", ledPath, target, err)
 		}
+	}
+	return nil
+}
+
+func setLedsToUserControlledMode(platform platform.Platform) error {
+	if err := setLeds(platform.Linux.StatusLeds, ledTriggerTarget, string(ledTriggerNone)); err != nil {
+		return err
+	}
+	if err := setLeds(platform.Linux.UserLeds, ledTriggerTarget, string(ledTriggerNone)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func restoreLedsState(platform platform.Platform) error {
+	if err := setLeds(platform.Linux.StatusLeds, ledBrightnessTarget, string("0")); err != nil {
+		return err
+	}
+	if err := setLeds(platform.Linux.UserLeds, ledBrightnessTarget, string("0")); err != nil {
+		return err
+	}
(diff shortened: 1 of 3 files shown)


internal/orchestrator/helpers.go
```diff
@@ -196,22 +196,55 @@ func GetCustomErrorFomDockerEvent(message string) error {
...
-	for _, ledPath := range platform.Linux.StatusLeds {
-		ledPath = ledPath.Join("trigger")
+type ledTrigger string
+
+const (
+	ledTriggerNone    ledTrigger = "none"
+	ledTriggerDefault ledTrigger = "default"
+)
+
+func setLeds(leds paths.PathList, target ledTarget, value string) error {
```
lucarin91: I would make this working to a single led
```suggestion
func setLed(leds *path.Path, target ledTarget, value string) error {
```

internal/orchestrator/helpers.go
```diff
@@ -196,22 +196,55 @@ func GetCustomErrorFomDockerEvent(message string) error {
...
 	return nil
 }
+
+func setLedsToUserControlledMode(platform platform.Platform) error {
+	if err := setLeds(platform.Linux.StatusLeds, ledTriggerTarget, string(ledTriggerNone)); err != nil {
+		return err
+	}
+	if err := setLeds(platform.Linux.UserLeds, ledTriggerTarget, string(ledTriggerNone)); err != nil {
+		return err
+	}
```
lucarin91: I would make eplict the fact that we are applying the same config on all the leds

```suggestion
    for _, led := range slices.Concat(platform.Linux.StatusLeds, platform.Linux.UserLeds) {
	    if err := setLed(led, ledTriggerTarget, none); err != nil {
		    return err
	    }
```

internal/orchestrator/helpers.go
```diff
@@ -196,22 +196,55 @@ func GetCustomErrorFomDockerEvent(message string) error {
...
+	if err := setLeds(platform.Linux.UserLeds, ledBrightnessTarget, string("0")); err != nil {
+		return err
+	}
+	if err := setLeds(platform.Linux.StatusLeds, ledTriggerTarget, string(ledTriggerDefault)); err != nil {
+		return err
+	}
+	if err := setLeds(platform.Linux.UserLeds, ledTriggerTarget, string(ledTriggerDefault)); err != nil {
+		return err
+	}
```
lucarin91: also here

```suggestion
    for _, led := range slices.Concat(platform.Linux.StatusLeds, platform.Linux.UserLeds) {
	    if err := setLeds(led, ledBrightnessTarget, "0"); err != nil {
		    return err
	    }
	    if err := setLeds(led, ledTriggerTarget, default); err != nil {
		    return err
	    }
	}
```

internal/orchestrator/helpers.go
```diff
@@ -196,22 +196,55 @@ func GetCustomErrorFomDockerEvent(message string) error {
...
 
-func setStatusLeds(platform platform.Platform, trigger LedTrigger) error {
-	for _, ledPath := range platform.Linux.StatusLeds {
-		ledPath = ledPath.Join("trigger")
+type ledTrigger string
+
+const (
+	ledTriggerNone    ledTrigger = "none"
+	ledTriggerDefault ledTrigger = "default"
+)
```
lucarin91: The `ledTrigger` type seems to just complicate the code, so I think it can be removed
```suggestion
const (
	none    string = "none"
	default string = "default"
)
```

lucarin91 (commented):
Good, so the issue was that we weren't setting the britness at 0. Good catch.

I would refactor the code as follows. What do you think?

lucarin91 (commented):
Also, if setting `default` to the user LEDs doesn't create any issue, I think we can also merge `StatusLeds` and `UserLeds` in the platform struct.

lucarin91:
> Also, if setting `default` to the user LEDs doesn't create any issue, I think we can also merge `StatusLeds` and `UserLeds` in the platform struct.

something like

```
		Leds: paths.NewPathList(
			// first 
			"/sys/class/leds/blue:bt",
			"/sys/class/leds/green:wlan",
			"/sys/class/leds/red:panic",
         		// second
	 		"/sys/class/leds/blue:user",
	 		"/sys/class/leds/green:user",
	 		"/sys/class/leds/red:user",
	 	),

```

internal/orchestrator/helpers.go
```diff
@@ -196,21 +196,45 @@ func GetCustomErrorFomDockerEvent(message string) error {
...
+		if err := setLeds(led, ledTriggerTarget, noneTrigger); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func restoreLedsState(platform platform.Platform) error {
+	for _, led := range platform.Linux.BoardLeds {
+		if err := setLeds(led, ledBrightnessTarget, string("0")); err != nil {
```
lucarin91, line 233:
```suggestion
		if err := setLeds(led, ledBrightnessTarget, "0"); err != nil {
```

internal/orchestrator/helpers.go
```diff
@@ -196,21 +196,45 @@ func GetCustomErrorFomDockerEvent(message string) error {
...
-	for _, ledPath := range platform.Linux.StatusLeds {
-		ledPath = ledPath.Join("trigger")
-		if !ledPath.Exist() {
-			return fmt.Errorf("LED path %s does not exist", ledPath)
+const (
+	noneTrigger    string = "none"
+	defaultTrigger string = "default"
+)
+
+func setLeds(ledPath *paths.Path, target ledTarget, value string) error {
```
lucarin91, line 211:
```suggestion
func setLed(ledPath *paths.Path, target ledTarget, value string) error {
```

lucarin91 (approved):

PR #369, opened 2026-04-23
Add a missing lock and defer lock acquisition.

martacarbone:
### Motivation

The current implementation contains return statements between the lock acquisition and the defer call, causing the resource to remain locked.
This PR moves the lock acquisition to just before the critical section to prevent lock leaks.

### Reviewer checklist

### DIFF
--- a/internal/update/arduino/arduino.go
+++ b/internal/update/arduino/arduino.go
@@ -177,6 +177,11 @@ func selectBestVersion(available []string, installed *semver.Version, constraint
 
 // UpgradePackages implements ServiceUpdater.
 func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, packages []update.PackageInfo, eventCB update.EventCallback) error {
+	if !a.lock.TryLock() {
+		return update.ErrOperationAlreadyInProgress
+	}
+	defer a.lock.Unlock()
+
 	if len(packages) == 0 {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


martacarbone:
Co-authored-by: Name <l.rindi@arduino.cc>

internal/update/arduino/arduino.go
cmaglie, line 194: This condition reads `a.platform`, should it be moved after the lock?

lucarin91: platform is only read, and never changes, so it should be safe

lucarin91: anyway maybe it is cleaner to move back the lock at the beginning

internal/update/arduino/arduino.go
```diff
@@ -177,10 +177,6 @@ func selectBestVersion(available []string, installed *semver.Version, constraint
 
 // UpgradePackages implements ServiceUpdater.
 func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, packages []update.PackageInfo, eventCB update.EventCallback) error {
-	if !a.lock.TryLock() {
-		return update.ErrOperationAlreadyInProgress
-	}
```
lucarin91, line 182: I would restore this code and add only the defer
```go
	defer a.lock.Unlock()
```

internal/update/arduino/arduino.go
```diff
@@ -196,6 +192,11 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, packages [
 		return fmt.Errorf("target version is empty for package '%s'", pkg.Name)
 	}
 
+	if !a.lock.TryLock() {
+		return update.ErrOperationAlreadyInProgress
+	}
+	defer a.lock.Unlock()
+
```
lucarin91:
```suggestion
```

lucarin91 (commented):
I think it this way it is more cler

cmaglie (approved):

internal/update/update.go
```diff
@@ -136,6 +133,10 @@ func (m *Manager) UpgradePackages(ctx context.Context, pkgs []UpgradablePackage)
 		}
 	}
 
+	if !m.lock.TryLock() {
+		return ErrOperationAlreadyInProgress
+	}
+
```
lucarin91, line 139: nit
```suggestion
```

lucarin91 (approved):

PR #374, opened 2026-04-27
Pass the information CONFIGURED_CARRIERS to the bricks container

mirkoCrobu:
### Motivation

The CONFIGURED_CARRIERS environment variable will pass the information to the containers/brick if any carrier (i.e. Media Carrier) has been configured by the user with the `arduino-linux-config` CLI.

Note: we do not really know if the carrier is connected, but the user has said it is.

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/linuxConfig/linuxConfig.go
+++ b/internal/orchestrator/linuxConfig/linuxConfig.go
@@ -28,8 +28,7 @@ import (
 
 const linuxConfigTool = "arduino-linux-config"
 
-func GetEnabledCarriers() ([]string, error) {
-
+func GetEnabledCarriers(ctx context.Context) ([]string, error) {
 	if _, err := exec.LookPath(linuxConfigTool); err != nil {
 		return nil, fmt.Errorf("arduino-linux-config tool not found in PATH: %w", err)
 	}
@@ -39,7 +38,7 @@ func GetEnabledCarriers() ([]string, error) {
 		return nil, fmt.Errorf("failed to create process 'arduino-linux-config carrier show': %w", err)
 	}
 
-	stdout, stderr, err := cmd.RunAndCaptureOutput(context.Background())
+	stdout, stderr, err := cmd.RunAndCaptureOutput(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to execute 'arduino-linux-config carrier show': %w\nstderr: %s", err, string(stderr))
 	}
--- a/internal/orchestrator/linuxConfig/linuxConfig.go
+++ b/internal/orchestrator/linuxConfig/linuxConfig.go
@@ -1,0 +1,53 @@
+// This file is part of arduino-app-cli.
+//
+// Copyright (C) Arduino s.r.l. and/or its affiliated companies
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package linuxconfig
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+)
+
+const linuxConfigTool = "arduino-linux-config"
+
+func CarrierShow() (*CarrierStatusOutput, error) {
+	if _, err := exec.LookPath(linuxConfigTool); err != nil {
+		return nil, fmt.Errorf("arduino-linux-config tool not found in PATH: %w", err)
+	}
+
+	cmd := exec.Command(linuxConfigTool, "carrier", "show", "--format", "json")
+
+	var out bytes.Buffer
+	var stderr bytes.Buffer
+	cmd.Stdout = &out
+	cmd.Stderr = &stderr
+
+	err := cmd.Run()
+	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/orchestrator/linuxConfig/linuxConfig.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+)
+
+const linuxConfigTool = "arduino-linux-config"
+
+func CarrierShow() (*CarrierStatusOutput, error) {
```
lucarin91: I would add a function that does exactly what we need, ie., the list of enabled carriers
```suggestion
func GetEnableCarriers() ([]string, error) {
```

internal/orchestrator/linuxConfig/linuxConfig.go
```diff
@@ -0,0 +1,53 @@
...
+	if _, err := exec.LookPath(linuxConfigTool); err != nil {
+		return nil, fmt.Errorf("arduino-linux-config tool not found in PATH: %w", err)
```
lucarin91, line 33: If `arduino-linux-config` isn't installed, I would just return an empty list of carriers
```suggestion
		return CarrierStatusOutput{}, nil
```

internal/orchestrator/linuxConfig/linuxConfig.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	cmd := exec.Command(linuxConfigTool, "carrier", "show", "--format", "json")
```
lucarin91: Can you use paths.NewProcess instead

internal/orchestrator/peripherals/devices.go
```diff
@@ -189,8 +191,23 @@ func HasVirtualDevice(deviceClass DeviceClass, devices []string) bool {
...
+		return nil, fmt.Errorf("unable to get configured carriers: %w", err)
+	}
+
+	var enabled []string
+	for _, c := range carriersStatus.Carriers {
+		if c.CurrentEnabled {
+			enabled = append(enabled, c.CarrierName)
+		}
+	}
+	return enabled, nil
```
lucarin91: we can remove all this function and the test

lucarin91 (changes requested):

internal/orchestrator/linuxConfig/linuxConfig.go
mirkoCrobu: Every error in the function is treated as a warning at a higher level. I added this for traceability. For every error, we return an empty string, but no errors are blocking.

Xayton (approved):

internal/orchestrator/linuxConfig/linuxConfig.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"os/exec"
+
+	"github.com/arduino/go-paths-helper"
+)
+
+const linuxConfigTool = "arduino-linux-config"
+
+func GetEnabledCarriers() ([]string, error) {
+
```
lucarin91: nit
```suggestion
```

internal/orchestrator/linuxConfig/linuxConfig.go
lucarin91: Yeah, but do we want to return a warning if arduino-linux-config isn't installed?

internal/orchestrator/linuxConfig/linuxConfig.go
```diff
@@ -0,0 +1,59 @@
...
+	if _, err := exec.LookPath(linuxConfigTool); err != nil {
+		return nil, fmt.Errorf("arduino-linux-config tool not found in PATH: %w", err)
+	}
+
+	cmd, err := paths.NewProcess(nil, linuxConfigTool, "carrier", "show", "--format", "json")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create process 'arduino-linux-config carrier show': %w", err)
+	}
+
+	stdout, stderr, err := cmd.RunAndCaptureOutput(context.Background())
```
lucarin91: I will take context from caller

```suggestion
	stdout, stderr, err := cmd.RunAndCaptureOutput(ctx)
```

internal/orchestrator/linuxConfig/linuxConfig.go
```diff
@@ -0,0 +1,59 @@
...
+	"encoding/json"
+	"fmt"
+	"os/exec"
+
+	"github.com/arduino/go-paths-helper"
+)
+
+const linuxConfigTool = "arduino-linux-config"
+
+func GetEnabledCarriers() ([]string, error) {
```
lucarin91:
```suggestion
func GetEnabledCarriers(ctx context.Context) ([]string, error) {
```

internal/orchestrator/orchestrator.go
```diff
@@ -313,9 +314,13 @@ func getAppEnvironmentVariables(app app.ArduinoApp, brickIndex *bricksindex.Bric
 		envs["VIDEO_DEVICE"] = videoDevices[0]
 	}
 
-	if mediaCarriers := peripherals.GetMediaCarriers(); len(mediaCarriers) > 0 {
-		envs["CONNECTED_CARRIERS"] = strings.Join(mediaCarriers, ",")
+	mediaCarriers, err := linuxconfig.GetEnabledCarriers()
+	if err != nil {
```
lucarin91:
```suggestion
	if mediaCarriers, err := linuxconfig.GetEnabledCarriers(context.Background()); err != nil {
```

lucarin91 (approved):

internal/orchestrator/linuxConfig/linuxConfig.go
mirkoCrobu: right.

internal/orchestrator/orchestrator.go
mirkoCrobu: @lucarin91 I implemented `mediaCarriers, err := linuxconfig.GetEnabledCarriers(ctx)`
making` getAppEnvironmentVariables` acepting a ctx, to propagate the context from `StartApp`

what do you think? Or should we keep it simpler as you suggested?

lucarin91: that's better, thanks

PR #375, opened 2026-04-27
Do not list examples incompatible with the current board (because using incompatible bricks)

cmaglie:
### Motivation

Do not list examples incompatible with the current board.

### Change description

Example apps are listed only if the required bricks are available (the bricks index loading already filter by board compatibility).

### Reviewer checklist

- [X] Changes are covered by tests.


lucarin91:
We should wait for a new example release so we can test the feature and verify that it correctly filters out not supported examples on UNO Q

internal/orchestrator/models.go
```diff
@@ -156,7 +156,7 @@ var (
 	ErrIncompleteImpulse   = errors.New("inpulse not ready for deployment")
```
mirkoCrobu:
```suggestion
	ErrIncompleteImpulse   = errors.New("impulse not ready for deployment")
```

lucarin91:
> We should wait for a new example release so we can test the feature and verify that it correctly filters out not supported examples on UNO Q

tested on unoq successfully

lucarin91 (approved):

PR #378, opened 2026-04-28
fix: update reconnection flow

### LINKED ISSUE #381: Inconsistent "ErrOperationAlreadyInProgress" when checking for updates
When checking for updates (ListUpgradablePackages) a specific error is returned if an operation is already in progress. The code returns this error both if there is already an upgrade in progress and if there is a check in progress.

This is causing a infinite update in the App Lab because it will try to do a check for updates first, and see if an upgrade is in progress. In that case it will then connect to the events endpoint. But if there are multiple check for updates pending, the specific return code is given anyway, and App Lab will wrongly think there is an update in progress (while there is just a check for updates). In this case, opening the events not only is useless but it will lock there forever, because it will never return a "upgrade finished" event.

The specific return status should only be returned if an upgrade in progress, multiple checks are fine (and are serialized anyway with a lock in the underlying service).

See also: https://github.com/arduino/arduino-app-cli/issues/382

lucarin91:
### Motivation

The system update has a set of inconsistencies that make the reconnection to an ongoing update buggy.

First of all, the `update/check` operation shouldn't lock the upgrade operation; multiple parties could call the check without getting an `OPERATION_IN_PROGRESS`. Indeed, this error means that an `update/apply` is in progress not a check. The upgrade operation is the real locking operation that should return an error in case another upgrade operation is in progress.

~Second, the `system/events` stream should immediately return if there isn't any upgrade operation in progress, which will prevent having a stale SSE event that never returns.~

### Change description

1. Remove the global lock and use instead an atomic, which allows to check the status of the upgrade without changing it
2. Make only the manager returning the `OPERATION_IN_PROGRESS` the two update services, only use a global lock.
~3. Substribe method returns a closed channel if no events will ever arrive~

### Reviewer checklist

### DIFF
--- a/internal/update/apt/service.go
+++ b/internal/update/apt/service.go
@@ -50,9 +50,7 @@ func New() *Service {
 // It filters the packages using the provided matcher function.
 // It returns a slice of UpgradablePackage or an error if the command fails.
 func (s *Service) ListUpgradablePackages(ctx context.Context, matcher func(update.UpgradablePackage) bool) ([]update.UpgradablePackage, error) {
-	if !s.lock.TryLock() {
-		return nil, update.ErrOperationAlreadyInProgress
-	}
+	s.lock.Lock()
 	defer s.lock.Unlock()
 
 	// Attempt to fix dpkg database in case an upgrade was interrupted in the middle.
@@ -76,10 +74,7 @@ func (s *Service) ListUpgradablePackages(ctx context.Context, matcher func(updat
 // It publishes events to subscribers during the upgrade process.
 // It returns an error if the upgrade is already in progress or if the upgrade command fails.
 func (s *Service) UpgradePackages(ctx context.Context, packages []update.PackageInfo, eventCB update.EventCallback) error {
-	if !s.lock.TryLock() {
-		return update.ErrOperationAlreadyInProgress
-	}
-
+	s.lock.Lock()
 	defer s.lock.Unlock()
 
 	// At the end of the upgrade, always try to restart the services (that need it).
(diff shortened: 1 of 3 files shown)


Xayton (commented):
Problems I see with the current changes:
1. changing the TryLock to Lock in the update services is probably not needed if the caller is making sure now concurrent operations are happening. And if we really want to be sure I would prefer to fail than locking (and waiting).
2. the atomic check is not the same as a lock. It will not prevent the critical code to be executed by multiple threads or requests. It's just thread-safe, but still affected by race conditions.

lucarin91:
>changing the TryLock to Lock in the update services is probably not needed if the caller is making sure now concurrent operations are happening. And if we really want to be sure I would prefer to fail than locking (and waiting).

We need that to avoid returning the `ErrOperationAlreadyInProgress` in the check operation. That error should be only returned if there is an update in progress. Also, in general, it is better not to fail during checks, as that should be a reliable operation.

>the atomic check is not the same as a lock. It will not prevent the critical code to be executed by multiple threads or requests. It's just thread-safe, but still affected by race conditions.

True, but we don't need to prevent datarace; the internal services already use locks, so we are safe that there will not be two concurrent updates. I swapped the lock with atomic, because in multiple places we need to check if there is an update running without grabbing the lock, which will cause spurious `ErrOperationAlreadyInProgress`.

mirkoCrobu:
**About ListUpgradablePackages** 
If we want to keep `ListUpgradablePackages` accessible from any client, we can remove any sync mechanism from `func (m *Manager) ListUpgradablePackages` and keep the lock for `func (s *Service) ListUpgradablePackages` and `func (a *ArduinoPlatformUpdater) ListUpgradablePackages`.

In this way, `ListUpgradablePackages` never fails even if an `UpgradePackages` is ongoing, and we can ensure that apt and platform are called one at a time.
But with a blocking `Lock()` in the services, if an `UpgradePackages` is ongoing, `ListUpgradablePackages` will wait until it finishes, which could take a while(is that what we want?).

mirkoCrobu:
About **UpgradePackages** and **Subscribe**
In both `UpgradePackages` and `Subscribe` the fix should be:

```
lock
isUpgrading check/swap
... 
unlock
```
in order that the critical operations are performed while holding the lock. 
In the current solution, there is a race condition since in both functions, the `isUpgrading check` and the `mu.Lock()` are not atomic and another goroutine can change the state in between.

lucarin91:
superseeded by https://github.com/arduino/arduino-app-cli/pull/380

PR #380, opened 2026-04-29
fix: update reconnection flow (mutex)

### LINKED ISSUE #381: Inconsistent "ErrOperationAlreadyInProgress" when checking for updates

lucarin91:
### Motivation

The system update has a set of inconsistencies that make the reconnection to an ongoing update buggy.

First of all, the `update/check` operation shouldn't lock the upgrade operation; multiple parties could call the check without getting an `OPERATION_IN_PROGRESS`. Indeed, this error means that an `update/apply` is in progress not a check. The upgrade operation is the real locking operation that should return an error in case another upgrade operation is in progress.

~Second, the `system/events` stream should immediately return if there isn't any upgrade operation in progress, which will prevent having a stale SSE event that never returns.~

### Change description

1. Remove the global lock and use instead an atomic, which allows to check the status of the upgrade without changing it
2. Make only the manager returning the `OPERATION_IN_PROGRESS` the two update services, only use a global lock.
~3. Substribe method returns a closed channel if no events will ever arrive~

### Reviewer checklist

### DIFF
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -24,7 +24,6 @@ import (
 	"net/http"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"golang.org/x/sync/errgroup"
@@ -58,7 +57,8 @@ type ServiceUpdater interface {
 }
 
 type Manager struct {
-	isUpgrading                  atomic.Bool
+	isUgradingMutex              sync.RWMutex
+	isUpgrading                  bool
 	debUpdateService             ServiceUpdater
 	arduinoPlatformUpdateService ServiceUpdater
 
@@ -75,9 +75,12 @@ func NewManager(debUpdateService ServiceUpdater, arduinoPlatformUpdateService Se
 }
 
 func (m *Manager) ListUpgradablePackages(ctx context.Context, matcher func(UpgradablePackage) bool) ([]UpgradablePackage, error) {
-	if m.isUpgrading.Load() {
+	m.isUgradingMutex.RLock()
+	if m.isUpgrading {
+		defer m.isUgradingMutex.RUnlock()
 		return nil, ErrOperationAlreadyInProgress
 	}
+	m.isUgradingMutex.RUnlock()
 
 	// Make sure to be connected to the internet, before checking for updates.
 	// This is needed because the checks below work also when offline (using cached data).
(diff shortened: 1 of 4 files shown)


internal/update/update.go
```diff
@@ -74,10 +75,12 @@ func NewManager(debUpdateService ServiceUpdater, arduinoPlatformUpdateService Se
 }
 
 func (m *Manager) ListUpgradablePackages(ctx context.Context, matcher func(UpgradablePackage) bool) ([]UpgradablePackage, error) {
-	if !m.lock.TryLock() {
+	m.isUgradingMutex.RLock()
+	if m.isUpgrading {
+		defer m.isUgradingMutex.RUnlock()
```
Xayton: The defer on this line is probably redundant.

internal/update/update.go
```diff
@@ -57,7 +57,8 @@ type ServiceUpdater interface {
 }
 
 type Manager struct {
-	lock                         sync.Mutex
+	isUgradingMutex              sync.RWMutex
```
Xayton:
```suggestion
	isUpgradingMutex              sync.RWMutex
```

internal/update/update.go
```diff
@@ -74,10 +75,12 @@ func NewManager(debUpdateService ServiceUpdater, arduinoPlatformUpdateService Se
 }
 
 func (m *Manager) ListUpgradablePackages(ctx context.Context, matcher func(UpgradablePackage) bool) ([]UpgradablePackage, error) {
-	if !m.lock.TryLock() {
+	m.isUgradingMutex.RLock()
```
Xayton:
```suggestion
	// If an upgrade is in progress, return a specific error so that the App Lab knows and can open the 
	// events stream. This is returned only if an upgrade is in progress, not if another check for updates
	// is running (which is acceptable).
	m.isUgradingMutex.RLock()
```

internal/update/update.go
```diff
@@ -74,10 +76,9 @@ func NewManager(debUpdateService ServiceUpdater, arduinoPlatformUpdateService Se
 }
 
 func (m *Manager) ListUpgradablePackages(ctx context.Context, matcher func(UpgradablePackage) bool) ([]UpgradablePackage, error) {
-	if !m.lock.TryLock() {
+	if m.isUpgrading.Load() {
```
Xayton, line 80:
```suggestion
	// Atomically check if an upgrade operation is already in progress. See https://github.com/arduino/arduino-app-cli/issues/381.
	if m.isUpgrading.Load() {
```

Xayton (approved):
Tested and it works (I could not reproduce the issue anymore).
Updates still work.

lucarin91:
The new tests highlight the following inconsistency in the upgrade events https://github.com/arduino/arduino-app-cli/issues/382

PR #393, opened 2026-05-07
feat: reuse port forwards

### LINKED ISSUE #14: Multiple random ADB port forwards are opened when the same app is run multiple times
## Motivation
If the same example is run multiple times, the system creates multiple adb forward rules pointing ot the same "internal" port 7000.

The forwarded ports are deleted only when the board is disconnected.

## How to reproduce
- clone the "Blink Led with UI" example
- first run: an adb forward is added  `7000`. Adb create  the following forward rules
```
❯ adb forward --list
4232352944 tcp:8800 tcp:8800
4232352944 tcp:7000 tcp:7000
```
- stop and run again: open the interface to a random port `6088`
```
❯ adb forward --list
4232352944 tcp:8800 tcp:8800
4232352944 tcp:7000 tcp:7000
4232352944 tcp:6088 tcp:7000
```
- stop and run again: open the interface to a random port `7177`
```
❯ adb forward --list
4232352944 tcp:8800 tcp:8800
4232352944 tcp:7000 tcp:7000
4232352944 tcp:6088 tcp:7000
4232352944 tcp:7177 tcp:7000
```

## Expected 
Remove the port forward 7000 when the app is stopped. 
The app is started on port 7000 again.
The random port are closed is the app is stopped.

## Possible solution
Remove the adb forward port when the app is stopped.
```
adb forward --remove tcp:8667
```

Xayton:
### Motivation
Check if a port forward request is already available. In that case the Forward function should reuse it, instead of opening a new one.

### Change description
- adb: we use adb forward --list to check if the requested port forward is already existing
- ssh: we track the currently opened port forwards, and check that list

### Additional Notes
The Forward function is used [here](https://github.com/bcmi-labs/cloud-editor-mono/blob/main/standalone-apps/app-lab-desktop/internal/tunnel/tunnel.go#L35) by the App Lab.

### DIFF
--- a/pkg/board/remote/adb/adb.go
+++ b/pkg/board/remote/adb/adb.go
@@ -110,21 +110,37 @@ func FromHost(host string, adbPath string) (*ADBConnection, error) {
 }
 
 func (a *ADBConnection) Forward(ctx context.Context, localPort int, remotePort int) error {
+	localString := fmt.Sprintf("tcp:%d", localPort)
+	remoteString := fmt.Sprintf("tcp:%d", remotePort)
+
 	if !ports.IsAvailable(localPort) {
+		// Check if the port is already forwarded by adb to the expected remote port
+		checkCmd, err := paths.NewProcess(nil, a.adbPath, "-s", a.host, "forward", "--list")
+		if err == nil {
+			if out, err := checkCmd.RunAndCaptureCombinedOutput(ctx); err == nil {
+				scanner := bufio.NewScanner(bytes.NewReader(out))
+				for scanner.Scan() {
+					// Output format is typically: <serial> <local> <remote>
+					fields := strings.Fields(scanner.Text())
+					if len(fields) >= 3 && fields[0] == a.host && fields[1] == localString && fields[2] == remoteString {
+						return nil // Port is already forwarded exactly as requested
+					}
+				}
+			}
+		}
+
 		return remote.ErrPortAvailable
 	}
 
-	local := fmt.Sprintf("tcp:%d", localPort)
-	remote := fmt.Sprintf("tcp:%d", remotePort)
-	cmd, err := paths.NewProcess(nil, a.adbPath, "-s", a.host, "forward", local, remote)
+	cmd, err := paths.NewProcess(nil, a.adbPath, "-s", a.host, "forward", localString, remoteString)
 	if err != nil {
 		return err
 	}
 	if out, err := cmd.RunAndCaptureCombinedOutput(ctx); err != nil {
 		return fmt.Errorf(
 			"failed to forward ADB port %s to %s: %w: %s",
-			local,
-			remote,
+			localString,
+			remoteString,
 			err,
 			out,
 		)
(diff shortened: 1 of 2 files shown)


pkg/board/remote/adb/adb.go
```diff
@@ -110,21 +110,37 @@ func FromHost(host string, adbPath string) (*ADBConnection, error) {
 }
 
 func (a *ADBConnection) Forward(ctx context.Context, localPort int, remotePort int) error {
+	localString := fmt.Sprintf("tcp:%d", localPort)
+	remoteString := fmt.Sprintf("tcp:%d", remotePort)
+
 	if !ports.IsAvailable(localPort) {
+		// Check if the port is already forwarded by adb to the expected remote port
+		checkCmd, err := paths.NewProcess(nil, a.adbPath, "-s", a.host, "forward", "--list")
+		if err == nil {
```
lucarin91:
```suggestion
		if checkCmd, err := paths.NewProcess(nil, a.adbPath, "-s", a.host, "forward", "--list"); err == nil {
```

pkg/board/remote/ssh/ssh.go
```diff
@@ -38,12 +38,18 @@ import (
 
 var ErrAuthFailed = errors.New("ssh authentication failed")
 
+type ForwardedPort struct {
+	Listener   net.Listener
+	LocalPort  int
+	RemotePort int
+}
+
```
lucarin91: I would move after the `SSHConnection` definition

pkg/board/remote/ssh/ssh.go
```diff
@@ -74,6 +80,18 @@ func FromHost(user, password, address string) (*SSHConnection, error) {
...
+	a.mu.Lock()
+	for _, fp := range a.ForwardedPorts {
+		if fp.LocalPort == localPort {
+			a.mu.Unlock()
+			if fp.RemotePort == remotePort {
+				return nil // Port already forwarded as requested
+			}
+			return remote.ErrPortAvailable // Already mapped but to a different remote port
+		}
+	}
```
lucarin91: Here I would only check if we forward the same port, the other check is done below in IsAvailable
```suggestion
    if slices.ContainsFunc(a.ForwardedPorts, func(p ForwardedPort) {
        return fp.LocalPort == localPort && fp.RemotePort == remotePort
    }) {
    	return nil // Port already forwarded as requested
	}
```

pkg/board/remote/ssh/ssh.go
```diff
@@ -74,6 +80,18 @@ func FromHost(user, password, address string) (*SSHConnection, error) {
 }
 
 func (a *SSHConnection) Forward(ctx context.Context, localPort int, remotePort int) error {
+	a.mu.Lock()
```
lucarin91, line 84: I would avoid possible issue by taking the lock for the whole method execution
```suggestion
	a.mu.Lock()
	defer a.mu.Unlock()
```

lucarin91 (approved):

pkg/board/remote/ssh/ssh.go
Xayton: yes it was this way because of the Unlock. Now that it's removed, this can be simplified.

PR #399, opened 2026-05-11
feat(pkg/board): add CopyFS api

lucarin91:
### Motivation

Add a simple api similar to https://pkg.go.dev/os#CopyFS for copying an FS into the board.

### Change description

This could be used both to copy files from the local filesystem to the board, and from the board itself.

```go
// from the local os to the board 
fs := os.DirFS("./a local path")
err := remote.CopyFS(conn, fs, "./a remote path")


// from the board to the board
fs := remotefs.New(conn, "./a path from the board")
err := remote.CopyFS(conn, fs, "./a remote path")
```

### Reviewer checklist

### DIFF
--- a/pkg/board/remote/fs.go
+++ b/pkg/board/remote/fs.go
@@ -1,0 +1,36 @@
+package remote
+
+import (
+	"fmt"
+	"io/fs"
+	"path"
+)
+
+// CopyFS copies the file system fsys into a remote directory dir using the provided RemoteConn.
+//
+// Copying stops at and returns the first error encountered.
+func CopyFS(conn RemoteConn, dir string, fsys fs.FS) error {
+	return fs.WalkDir(fsys, ".", func(p string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		newPath := path.Join(dir, p)
+		switch d.Type() {
+		case fs.ModeDir:
+			return conn.MkDirAll(newPath)
+		case fs.ModeSymlink:
+			// FIXME: implement symbolic link support
+			return fmt.Errorf("CopyFS: symbolic links not supported")
+		case 0:
+			r, err := fsys.Open(p)
+			if err != nil {
+				return err
+			}
+			defer r.Close()
+			// FIXME: support permission fowarding
+			return conn.WriteFile(r, newPath)
+		default:
+			return &fs.PathError{Op: "CopyFS", Path: p, Err: fs.ErrInvalid}
+		}
+	})
+}
--- a/pkg/board/remote/fs.go
+++ b/pkg/board/remote/fs.go
@@ -1,3 +1,8 @@
+// This file is part of arduino-app-cli.
+//
+// SPDX-FileCopyrightText: Arduino s.r.l. and/or its affiliated companies
+// SPDX-License-Identifier: GPL-3.0-or-later
+
 package remote
 
 import (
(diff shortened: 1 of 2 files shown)


pkg/board/remote/fs.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"io/fs"
+	"path"
+)
+
+// CopyFS copies the file system fsys into a remote directory dir using the provided RemoteConn.
+//
+// Copying stops at and returns the first error encountered.
+func CopyFS(conn RemoteConn, dir string, fsys fs.FS) error {
```
dido18, line 17: suggestsion. `src` or `local` and `dst` or `remote` ?
```suggestion
func CopyFS(conn RemoteConn, src fs.FS, dst string) error {
```

pkg/board/remote/fs.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+		newPath := path.Join(dir, p)
+		switch d.Type() {
+		case fs.ModeDir:
+			return conn.MkDirAll(newPath)
+		case fs.ModeSymlink:
+			// FIXME: implement symbolic link support
+			return fmt.Errorf("CopyFS: symbolic links not supported")
+		case 0:
+			r, err := fsys.Open(p)
```
dido18, line 30:
```suggestion
			r, err := src.Open(p)
```

pkg/board/remote/fs.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+	"io/fs"
+	"path"
+)
+
+// CopyFS copies the file system fsys into a remote directory dir using the provided RemoteConn.
+//
+// Copying stops at and returns the first error encountered.
+func CopyFS(conn RemoteConn, dir string, fsys fs.FS) error {
+	return fs.WalkDir(fsys, ".", func(p string, d fs.DirEntry, err error) error {
```
dido18, line 18:
```suggestion
	return fs.WalkDir(src, ".", func(p string, d fs.DirEntry, err error) error {
```

pkg/board/remote/fs.go
```diff
@@ -0,0 +1,41 @@
...
+			return err
+		}
+		newPath := path.Join(dir, p)
+		switch d.Type() {
+		case fs.ModeDir:
+			return conn.MkDirAll(newPath)
+		case fs.ModeSymlink:
+			// FIXME: implement symbolic link support
+			return fmt.Errorf("CopyFS: symbolic links not supported")
+		case 0:
```
dido18, line 29: what is case 0 ?

dido18 (changes requested):

pkg/board/remote/fs.go
lucarin91: I would prefer to keep as close as possible to the os one https://pkg.go.dev/os#CopyFS

pkg/board/remote/fs.go
lucarin91: same

pkg/board/remote/fs.go
lucarin91: same

lucarin91: BTW, a [fs.FS](https://pkg.go.dev/io/fs#FS) is a read-only object, so it shouldn't be a destination

pkg/board/remote/fs.go
lucarin91: ask the golang developer 😄 . It seems the way to check if a node is a regular file. https://cs.opensource.google/go/go/+/refs/tags/go1.26.3:src/os/dir.go;l=166

dido18 (commented):
I would add unitest for the CopyFS , with any RemoteConn

PR #401, opened 2026-05-12
Board default model

martacarbone:
### Motivation

There is the need to define a default model per architecture/board

### Change description

This code adds in bricks-list.yaml the following per board model definition:
```
  model_by_boards:
    - board: ventunoq
      model_name: yolox-qnn-object-detection
    - board: unoq
      model_name: yolox-object-detection
```

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/bricks/bricks.go
+++ b/internal/orchestrator/bricks/bricks.go
@@ -22,6 +22,7 @@ import (
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/bricksindex"
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/config"
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/modelsindex"
+	"github.com/arduino/arduino-app-cli/internal/platform"
 )
 
 var (
@@ -62,7 +63,7 @@ func (s *Service) List() (BrickListResult, error) {
 	return res, nil
 }
 
-func (s *Service) AppBrickInstancesList(a *app.ArduinoApp) (AppBrickInstancesResult, error) {
+func (s *Service) AppBrickInstancesList(a *app.ArduinoApp, platform platform.Platform) (AppBrickInstancesResult, error) {
 	res := AppBrickInstancesResult{BrickInstances: make([]BrickInstance, len(a.Descriptor.Bricks))}
 	for i, brickInstance := range a.Descriptor.Bricks {
 		brick, found := s.bricksIndex.WithAppBricks(a.LocalBricks).FindBrickByID(brickInstance.ID)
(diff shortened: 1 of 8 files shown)


internal/orchestrator/bricks/bricks.go
```diff
@@ -100,6 +101,19 @@ func (s *Service) AppBrickInstancesList(a *app.ArduinoApp) (AppBrickInstancesRes
 	return res, nil
 }
 
+func getModelName(defaultModelName string, modelsBoard []bricksindex.ModelsBoard, boardName string) string {
```
mirkoCrobu:
```suggestion
func getModelNameByBoard(defaultModelName string, modelsBoard []bricksindex.ModelsBoard, boardName string) string {
```

internal/orchestrator/bricks/bricks_test.go
```diff
@@ -1084,3 +1084,66 @@ func TestLocalBrickRename(t *testing.T) {
...
+					ModelName: "ventunoq-model",
+				},
+			},
+			expectedModelName: "ventunoq-model",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			modelName := getModelName(tt.definedModelName, tt.definedModelByBoard, tt.boardName)
```
mirkoCrobu: What happens if tt.definedModelByBoard is empty?

internal/orchestrator/bricks/bricks.go
```diff
@@ -84,7 +85,7 @@ func (s *Service) AppBrickInstancesList(a *app.ArduinoApp) (AppBrickInstancesRes
 			Category:        brick.Category,
 			Status:          "installed",
 			RequireModel:    brick.RequireModel,
-			ModelID:         cmp.Or(brickInstance.Model, brick.ModelName),
+			ModelID:         cmp.Or(brickInstance.Model, getModelName(brick.ModelName, brick.ModelByBoard, platform.BoardName)),
```
mirkoCrobu: Do you think it could work if getModelName is a method of brick since 2 paramethers of 3 depends of bricks? 
```suggestion
			ModelID:         cmp.Or(brickInstance.Model, brick.GetModelName(platform.BoardName)),
```

internal/orchestrator/bricksindex/bricks_index.go
```diff
@@ -62,6 +62,11 @@ type BrickVariable struct {
 	Secret       bool   `yaml:"secret"`
 }
 
+type ModelsBoard struct {
+	Board     string `yaml:"board"`
```
mirkoCrobu, line 66: from where we retrieve this field? from the brick-index.yaml?

internal/orchestrator/bricks/bricks_test.go
martacarbone: The case is covered by test named "Brick with one default model".
I added a test with the nil slice, but this is the same of an empty slice

internal/orchestrator/bricksindex/bricks_index.go
martacarbone: It is added in the brick list according this definition: https://github.com/arduino/private-tooling-team/issues/506#issuecomment-4419036903

mirkoCrobu (approved):

PR #404, opened 2026-05-12
add progress info to the update flow

mirkoCrobu:
### Motivation

When launching the update process (both via API and, eventually, a CLI command), we should return a progress in percentage. This could be done in many ways:

estimate the time needed for some tasks, based on previous measurements (for example, how much it usually takes to do an apt update on Imola)
When we know there is a list of tasks to do (a list of packages to update), then it's possible to use that to provide progress information
This information can then be used by the FE team (App Lab) to show a progress bar.

### Reviewer checklist


internal/helpers/helpers.go
```diff
@@ -18,24 +18,30 @@ func ArduinoCLIDownloadProgressToString(progress *rpc.DownloadProgress) string {
...
 	}
+
 	if progress.GetCompleted() {
-		data += " completed"
-	} else {
-		data += fmt.Sprintf(" %.2f%%", progress.GetPercent())
+		return fmt.Sprintf("%s: completed", name)
 	}
-	return data
+	return fmt.Sprintf("%s: %.0f%%", name, progress.GetPercent())
```
MatteoPologruto, line 49: Originally it was printing `%.2f%%`. Is this change wanted?

internal/update/arduino/arduino.go
```diff
@@ -181,15 +181,36 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, packages [
...
+	makeDownloadProgressCallback := func(name string, basePercentage, phaseWeight float32) func(*rpc.DownloadProgress) {
+		return func(curr *rpc.DownloadProgress) {
+			data := helpers.ArduinoCLIDownloadProgressToString(curr)
+			eventCB(update.NewDataEvent(update.UpgradeLineEvent, data))
+			if updateInfo := curr.GetUpdate(); updateInfo != nil {
+				if updateInfo.GetTotalSize() <= 0 {
+					return
+				}
+				localProgress := (float32(updateInfo.GetDownloaded()) / float32(updateInfo.GetTotalSize())) * 100.0
+				totalArduinoProgress := basePercentage + (localProgress/100.0)*phaseWeight
```
MatteoPologruto:
```suggestion
				localProgress := float32(updateInfo.GetDownloaded()) / float32(updateInfo.GetTotalSize())
				totalArduinoProgress := basePercentage + localProgress*phaseWeight
```
I think this can be simplified

internal/helpers/helpers.go
mirkoCrobu: yes, i through that 45% is better than 45.00% or 45.67%. What do you think?

internal/update/event.go
```diff
@@ -14,16 +14,22 @@ const (
...
+	Type     EventType
+	progress *ProgressInfo
+	data     string
+	err      error // error field for error events
+}
 
-	data string
-	err  error // error field for error events
+type ProgressInfo struct {
+	Name     string
```
Xayton: Name is a bit generic. What do you think about `Step` or `Phase`?

internal/helpers/helpers.go
Xayton: I think no decimal part is acceptable.

internal/update/event.go
```diff
@@ -14,16 +14,22 @@ const (
 	UpgradeLineEvent EventType = iota
 	StartEvent
 	RestartEvent
+	ProgressEvent
```
Xayton, line 17: This new event type should also be documented in the `internal/api/docs/openapi.yaml:1326` file.

internal/update/arduino/arduino.go
```diff
@@ -181,15 +181,36 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, packages [
...
 	}
 
-	downloadProgressCB := func(curr *rpc.DownloadProgress) {
-		data := helpers.ArduinoCLIDownloadProgressToString(curr)
-		slog.Debug("Download progress", slog.String("download_progress", data))
-		eventCB(update.NewDataEvent(update.UpgradeLineEvent, data))
+	const indexBase float32 = 0.0
+	const indexWeight float32 = 30.0
+	const upgradeBase float32 = 30.0
+	const upgradeWeight float32 = 60.0
```
Xayton: Weight of the arduino platform phase adds up to 90 instead of 100.

internal/update/apt/service.go
```diff
@@ -96,6 +96,7 @@ func (s *Service) UpgradePackages(ctx context.Context, packages []update.Package
 		return pkg.Name
 	})
 	eventCB(update.NewDataEvent(update.StartEvent, "Upgrade is starting"))
+	eventCB(update.NewProgressEvent("apt upgrade", 65.0))
```
Xayton: Is it correct that this starts at 65%?
Shouldn't this go from 0 to 100, as it is then later remapped to 20-100 by the `aptWeight` factor.

internal/update/arduino/arduino.go
```diff
@@ -181,15 +181,36 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, packages [
...
+				eventCB(update.NewProgressEvent(name, totalArduinoProgress))
+			}
+		}
 	}
-	taskProgressCB := func(msg *rpc.TaskProgress) {
-		data := helpers.ArduinoCLITaskProgressToString(msg)
-		slog.Debug("Task progress", slog.String("task_progress", data))
-		eventCB(update.NewDataEvent(update.UpgradeLineEvent, data))
+
+	makeTaskProgressCallback := func(name string, basePercentage, phaseWeight float32) func(*rpc.TaskProgress) {
```
Xayton: Looks like in this function the TaskProgress.Percent is always 0, because the arduino-cli PlatformInstall code does not set it. So this code will just output a progress of basePercentage (30%).

internal/update/arduino/arduino.go
```diff
@@ -215,19 +236,24 @@ func (a *ArduinoPlatformUpdater) UpgradePackages(ctx context.Context, packages [
...
 
+	platformDownloadCB := makeDownloadProgressCallback("platform download", upgradeBase, upgradeWeight)
+	platformTaskCB := makeTaskProgressCallback("platform upgrade", upgradeBase, upgradeWeight)
+
 	stream := commands.PlatformInstallStreamResponseToCallbackFunction(
 		ctx,
-		downloadProgressCB,
-		taskProgressCB,
+		platformDownloadCB,
+		platformTaskCB,
```
Xayton: It looks like the way arduino-cli works, the download and task phases are interleaved (even if we install just one platform, it is still composed by multiple downloads, like tools or libraries, and multiple install steps).

This breaks because the two callbacks for download and task assume they will happen in strict order (download first and then task). If they are interlevaed, the progress will jump un and down between 30 and 90.

- `makeDownloadProgressCallback` uses `downloaded/total_size` — real data, but *per-file*, so it resets to the band floor on each new archive (11 tools + platform for `arduino:zephyr@0.56.0`).
- `makeTaskProgressCallback` uses `msg.GetPercent()` — a field that is **always 0** here, so every narration line ("Downloading packages", "Installing platform …", "Configuring platform.") emits a flat `upgradeBase` = 30%.

internal/update/event.go
mirkoCrobu: done

internal/update/event.go
mirkoCrobu: done

internal/update/apt/service.go
mirkoCrobu: Right done

internal/update/arduino/arduino.go
mirkoCrobu: done

internal/update/arduino/arduino.go
mirkoCrobu: done

internal/update/arduino/arduino.go
mirkoCrobu: fixed

Xayton (approved):

PR #407, opened 2026-05-15
fix: some boards require an always running micro

lucarin91:
### Motivation

On Monza, the microcontroller runs some other task other than the sketch that we cannot stop. So we need to stop only the user sketch and not the other services.

### Change description

Instead of keeping the reset pin down to stop the micro, we can do a reset and make the sketch wait for the wait-for-app signal.

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -292,10 +292,9 @@ func stopAppWithCmd(ctx context.Context, docker command.Cli, platform platform.P
 		}
 		if running != nil && running.FullPath.String() == app.FullPath.String() {
 			cb(StreamMessage{data: "Stopping microcontroller..."})
-			if err := platform.GetMicro().Disable(); err != nil {
-				return err
-				// XXX: if we fail to stop the sketch, do we want to continue to stop the app anyway?
-				//      maybe we can just log the error and continue
+			// Reset the microcontroller stops the sketch because the micro will wait for the wait-for-app signal.
+			if err := platform.GetMicro().Reset(); err != nil {
+				slog.Warn("failed to reset microcontroller", slog.String("error", err.Error()))
 			}
 		}
 	}
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -292,7 +292,12 @@ func stopAppWithCmd(ctx context.Context, docker command.Cli, platform platform.P
 		}
 		if running != nil && running.FullPath.String() == app.FullPath.String() {
 			cb(StreamMessage{data: "Stopping microcontroller..."})
-			// Reset the microcontroller stops the sketch because the micro will wait for the wait-for-app signal.
+// Reset() triggers a reset of the microcontroller.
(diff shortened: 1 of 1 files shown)


internal/orchestrator/orchestrator.go
```diff
@@ -292,10 +292,9 @@ func stopAppWithCmd(ctx context.Context, docker command.Cli, platform platform.P
 		}
 		if running != nil && running.FullPath.String() == app.FullPath.String() {
 			cb(StreamMessage{data: "Stopping microcontroller..."})
-			if err := platform.GetMicro().Disable(); err != nil {
-				return err
-				// XXX: if we fail to stop the sketch, do we want to continue to stop the app anyway?
-				//      maybe we can just log the error and continue
+			// Reset the microcontroller stops the sketch because the micro will wait for the wait-for-app signal.
```
dido18: this is true only if the sketch has been compilted with the `unoq.menu.wait_linux_boot.app=Wait for App`, isn't ?

In the other cases 
- `unoq.menu.wait_linux_boot.yes=Wait for Linux` , the sketrch reset and wait the linux to start ? it blocks forever ?
- `unoq.menu.wait_linux_boot.no=Immediate` te sketch is reset and start again without waiting.

lucarin91: In both cases the sketch will just restart

dido18:
```suggestion
// Reset() triggers a reset of the microcontroller.
// Note: If the sketch was compiled with 'Wait for App', the microcontroller 
// will stop and wait for the wait-for-app signal before restarting the sketch. 
// In all other boot modes (e.g., 'Wait for Linux' or 'Immediate'), the 
// sketch will simply restart.

```

dido18 (approved):

internal/orchestrator/orchestrator.go
```diff
@@ -292,10 +292,14 @@ func stopAppWithCmd(ctx context.Context, docker command.Cli, platform platform.P
...
 			cb(StreamMessage{data: "Stopping microcontroller..."})
-			if err := platform.GetMicro().Disable(); err != nil {
-				return err
-				// XXX: if we fail to stop the sketch, do we want to continue to stop the app anyway?
-				//      maybe we can just log the error and continue
+// Reset() triggers a reset of the microcontroller.
+// Note: If the sketch was compiled with 'Wait for App', the microcontroller 
+// will stop and wait for the wait-for-app signal before restarting the sketch. 
+// In all other boot modes (e.g., 'Wait for Linux' or 'Immediate'), the 
+// sketch will simply restart.
```
dido18:
```suggestion
 				// Reset() triggers a reset of the microcontroller.
				// Note: If the sketch was compiled with 'Wait for App', the microcontroller 
				// will stop and wait for the wait-for-app signal before restarting the sketch. 
				// In all other boot modes (e.g., 'Wait for Linux' or 'Immediate'), the 
				// sketch will simply restart.
```

dido18 (approved):

PR #415, opened 2026-05-20
fix: e2e test timeout

Xayton:
Reduce the number of container images downloaded, by limiting the test to a "unoq" board.
This should reduce the timeout errors in the e2e tests.

Test run here: https://github.com/arduino/arduino-app-cli/actions/runs/26164192467

### DIFF
--- a/internal/e2e/updatetest/helpers.go
+++ b/internal/e2e/updatetest/helpers.go
@@ -7,7 +7,6 @@ package updatetest
 
 import (
 	"bufio"
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


internal/e2e/updatetest/helpers.go
```diff
@@ -227,6 +226,16 @@ func runSystemUpdate(t *testing.T, containerName string) {
 	require.NoError(t, err, "system update failed")
 }
 
+func removeDockerImage(t *testing.T, imageName string) {
+	t.Helper()
+
+	cmd := exec.Command("docker", "rmi", "-f", imageName)
+	fmt.Println("🧹 Removing Docker image " + imageName)
```
mirkoCrobu: Why not use t.Logf() or t.Log() instead of fmt.Println?

Should we avoid using the emoji?

Xayton: I copied what was in stopDockerContainer, which already had the emoji, but I agree.
I'm also updating the code to use `t.Logf()`, as it's cleaner and does not interleaves with other test messages.

lucarin91 (approved):

internal/e2e/updatetest/helpers.go
Xayton: Changes done.

PR #417, opened 2026-05-20
feat: parametric EI device type based on board platform

giulio93:
## Summary

- `InstallEIModel` now accepts a `platform.Platform` parameter and selects `deviceType` via a switch on `platform.BoardName` instead of a hardcoded value
- UNO Q → `runner-linux-aarch64`; VentunoQ → `runner-linux-aarch64-qnn` (Qualcomm Neural Network acceleration)
- Threads the platform through `HandleInstallEIModel` handler and the API router

Closes #326

### DIFF
--- a/internal/orchestrator/models.go
+++ b/internal/orchestrator/models.go
@@ -231,18 +231,13 @@ func isModelInUse(ctx context.Context, modelsIndex *modelsindex.ModelsIndex, doc
 
 func InstallEIModel(ctx context.Context, bricksIndex *bricksindex.BricksIndex, modelsIndex *modelsindex.ModelsIndex, dockerClient command.Cli, eiClient *edgeimpulse.EIClient, modelsDir *paths.Path, platform platform.Platform, projectID int, impulseID int) (AIModelItem, error) {
 
-	mType := "float32"
-	mEngine := "tflite"
-	var deviceType string
-	switch platform.BoardName {
-	case "unoq":
-		deviceType = "runner-linux-aarch64"
-	case "ventunoq":
-		deviceType = "runner-linux-aarch64-qnn"
+	eiParams, err := platform.EIDeploymentParams()
+	if err != nil {
+		return AIModelItem{}, err
 	}
 
 	id := fmt.Sprintf("ei-model-%d-%d", projectID, impulseID)
-	err := isModelInUse(ctx, modelsIndex, dockerClient, id)
+	err = isModelInUse(ctx, modelsIndex, dockerClient, id)
 	if err != nil {
 		return AIModelItem{}, fmt.Errorf("cannot install EI model: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


internal/orchestrator/models.go
```diff
@@ -258,13 +258,17 @@ func isModelInUse(ctx context.Context, modelsIndex *modelsindex.ModelsIndex, doc
 	return nil
 }
 
-func InstallEIModel(ctx context.Context, bricksIndex *bricksindex.BricksIndex, modelsIndex *modelsindex.ModelsIndex, dockerClient command.Cli, eiClient *edgeimpulse.EIClient, modelsDir *paths.Path, projectID int, impulseID int) (AIModelItem, error) {
+func InstallEIModel(ctx context.Context, bricksIndex *bricksindex.BricksIndex, modelsIndex *modelsindex.ModelsIndex, dockerClient command.Cli, eiClient *edgeimpulse.EIClient, modelsDir *paths.Path, platform platform.Platform, projectID int, impulseID int) (AIModelItem, error) {
 
-	// TODO these parameters aim to build a model optimized for the Imola hardware, they should change based on the target device
 	mType := "float32"
 	mEngine := "tflite"
```
mirkoCrobu: What about `myType `and `mEngine`? Will they be the same for `unoq `and `ventunoq`? 
Could we accept other values than `float32 `and `tflite`?

internal/orchestrator/models.go
```diff
@@ -258,13 +258,17 @@ func isModelInUse(ctx context.Context, modelsIndex *modelsindex.ModelsIndex, doc
...
-	deviceType := "runner-linux-aarch64"
-	var mversion int
+	var deviceType string
+	switch platform.BoardName {
+	case "unoq":
+		deviceType = "runner-linux-aarch64"
+	case "ventunoq":
+		deviceType = "runner-linux-aarch64-qnn"
+	}
```
mirkoCrobu: do we need a default value here? Could it be empty?

lucarin91: I would prefer to add a new field in the Platform struct called `EdgeImpulseTarget` for instance. In this way we will remember to add a valid value whenever we will add a new supported platform

giulio93: There is no default value it depends on the Platform type

giulio93: Soon we will integrate new 3rd parties model hub, i don't like to have for each party specific variable in the platfrom

lucarin91: We don't, those model providers will not include a custom builder job, so they don't need to specify their target, and also, if they will, we should keep all the specific board configuration in a single place, so we will not forget about them, if we add in the future other boards.

lucarin91: We should probably return an error then

internal/platform/platform.go
```diff
@@ -95,6 +96,23 @@ func (p Platform) SupportFlashToRam() bool {
...
+}
+
+func (p Platform) EIDeploymentParams() (EIDeploymentParams, error) {
+	switch p.BoardName {
+	case "unoq":
+		return EIDeploymentParams{ModelType: "float32", Engine: "tflite", DeviceType: "runner-linux-aarch64"}, nil
+	case "ventunoq":
+		return EIDeploymentParams{ModelType: "float32", Engine: "tflite", DeviceType: "runner-linux-aarch64-qnn"}, nil
+	default:
+		return EIDeploymentParams{}, fmt.Errorf("unsupported platform %q for Edge Impulse deployment", p.BoardName)
```
dido18, line 112: We could panic here, and remove the error as the result of the function

dido18 (approved):

PR #422, opened 2026-05-27
feat(pkg/board): improve FindAdbPath function

lucarin91:
### Motivation

Improve the mechanism by which we attempt to find the adb installed by arduino-cli.

### Change description

1. cache `FindAdbPath` result 
2. Use arduino-cli config and glob to find a real path, also in case of adb update

### Reviewer checklist


lucarin91:
- [x] we should remove the test that will not work in ci

mirkoCrobu (approved):

PR #424, opened 2026-05-27
feat/models download flow

giulio93:
### Motivation
app-bricks-py#227 introduced a new containerized models-downloader brick that can fetch and install AI models from Arduino AI Hub, Hugging Face, and Edge Impulse, with per-model progress reporting.

### Change description

**modelsindex**: extended the ModelsIndex and AIModel structs to carry platform-specific deployment metadata (handler, download location, supported boards). GetModels now takes a context and Docker client to query installed status at runtime by invoking the models-downloader container via list_models.py.

API routing: added two new HTTP endpoints:

PUT /v1/models/{modelID}/install — triggers a model download via the models-downloader Docker container and streams progress as SSE events.

**Assets**: bumped bundled brick assets from 0.10.1 to 0.11.0rc3, which includes the new models-downloader image.

### Additional Notes

Related PR: arduino/app-bricks-py#227


internal/orchestrator/modelsindex/models_index.go
```diff
@@ -38,6 +38,11 @@ func (b *assetsModelList) UnmarshalYAML(unmarshal func(any) error) error {
 	return nil
 }
 
+type ModelDeployment struct {
+	Handler   string `yaml:"handler"`
+	PreLoaded bool   `yaml:"pre-loaded"`
```
dido18: move after the AiModel definition

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -49,6 +54,8 @@ type AIModel struct {
 	Metadata          map[string]string `yaml:"metadata,omitempty"`
 	IsInternal        bool              `yaml:"-"`
 	SupportedBoards   []string          `yaml:"supported_boards,omitempty"`
+	DownloadLocation  string            `yaml:"download_location,omitempty"`
```
dido18: What is this download_location ?

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -38,6 +38,11 @@ func (b *assetsModelList) UnmarshalYAML(unmarshal func(any) error) error {
 	return nil
 }
 
+type ModelDeployment struct {
+	Handler   string `yaml:"handler"`
```
dido18: the array of platforms missing

For example:
```
     platforms:
        - ventunoq:
            variables:
              model_type: "voice_ai"
              model_name: "melotts_zh"
              models_repository: "models/audio-analytics/tts"
              model_directory: "melotts_zh-voice_ai-mixed_with_float-qualcomm_qcs8275"
              quantization: "mixed_with_float"
              chipset: "qualcomm-qcs8275"
              version: "0.51.0"
```

internal/api/api.go
```diff
@@ -62,9 +63,11 @@ func NewHTTPRouter(
 	mux.Handle("PUT /v1/system/update/apply", handlers.HandleUpdateApply(updater))
 	mux.Handle("GET /v1/system/resources", handlers.HandleSystemResources(cfg))
 
-	mux.Handle("GET /v1/models", handlers.HandleModelsList(modelsIndex))
-	mux.Handle("GET /v1/models/{modelID}", handlers.HandlerModelByID(modelsIndex))
+	mux.Handle("GET /v1/models", handlers.HandleModelsList(modelsIndex, cfg, platform))
+	mux.Handle("GET /v1/models/{modelID}", handlers.HandlerModelByID(modelsIndex, cfg))
 	mux.Handle("PUT /v1/models/ei/projects/{projectID}", handlers.HandleInstallEIModel(cfg, bricksIndex, modelsIndex, dockerClient))
+	mux.Handle("PUT /v1/models/{modelID}/install", handlers.HandleInstallModel(cfg, modelsIndex, installMgr))
+	mux.Handle("GET /v1/models/events", handlers.HandleModelInstallEvents(installMgr))
```
dido18: After a discussion with @lucarin91 we agreed on having a single endpoint that returns the SSE event of the download. If the client close the connection, the download is stopeed.

`PUT /v1/models/{modelID}`

The endpoint does the following
 - check if the model is already download, return 200
 - download the file and stream the progressvia SSE events
 - if the clients stops the sse connection, stop the download.

internal/orchestrator/model_install_manager.go
```diff
@@ -0,0 +1,62 @@
...
+
+// ModelInstallManager is a global pub/sub hub for model install progress events.
+// All subscribers receive events for every model; each event carries its ModelID
+// so the frontend can filter by model.
+type ModelInstallManager struct {
+	mu   sync.Mutex
+	subs map[chan ModelInstallEvent]struct{}
+}
+
+func NewModelInstallManager() *ModelInstallManager {
```
dido18: I do not think we need the install Manager, we could start simple wit a callaback to send event via sse

internal/orchestrator/models.go
```diff
@@ -43,35 +45,68 @@ type AIModelItem struct {
...
+		entries, err := runListAction(ctx, handler, cfg, plat)
+		if err != nil {
+			slog.Warn("cannot list models from handler", "handler", handler.ID, "err", err)
+			continue
+		}
+		for _, entry := range entries {
+			if existing, ok := resultMap[entry.ID]; ok {
+				existing.Installed = entry.Installed
+				resultMap[entry.ID] = existing
+			}
```
dido18: I think we need to move the logicin inside the `modelsindex.ListModels` and use the 

```
docker -v xxx//path/to/models-list.yaml run ${DOCKER_REGISTRY_BASE:-ghcr.io/arduino/}app-bricks/models-downloader:0.10.0 /app/list_models.py --model-list /path/to/models-list.yaml

```

.gitignore
```diff
@@ -37,3 +37,5 @@ build/
 # debian packaging files
 /debian/arduino-app-cli/home/arduino/.local/share/arduino-app-cli/examples
 /debian/arduino-app-cli/var/lib/arduino-app-cli/examples
+
+.mario/
```
dido18: Who is mario ? :)

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -59,10 +85,33 @@ type BrickConfig struct {
...
-func (m *ModelsIndex) GetModels() []AIModel {
-	return m.loadModels()
+func (m *ModelsIndex) GetModels(ctx context.Context, cli client.APIClient, cfg config.Configuration, plat platform.Platform) []AIModel {
+	models := m.loadModels()
+	statuses := m.Handlers.GetModelStatus(ctx, cli, cfg, plat)
+	for i := range models {
+		if installed, ok := statuses[models[i].ID]; ok {
+			models[i].Installed = installed
+		} else {
+			models[i].Installed = true
```
dido18: if the model is not found in the filesystem I think we must set installed=false

internal/dockerhelper/dockerhelper.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Run creates, starts, and waits for a container to exit, streaming stdout and
+// stderr to the provided writers. The container is always removed on return.
+//
+// ContainerLogs with Follow:true is used instead of ContainerAttach because
+// the logging driver fully buffers output before streaming it — fast-exiting
+// containers can close the raw attach connection mid-frame, producing spurious
+// "short write" errors.
+func Run(ctx context.Context, cli client.APIClient, opts RunOptions) error {
```
dido18, line 40: I do not think we need this wrapper docker handler

internal/orchestrator/modelsindex/models_index.go
giulio93: Is the volume mounted to the docker where the model will be downloaded

internal/orchestrator/modelsindex/models_index.go
giulio93: added https://github.com/arduino/arduino-app-cli/pull/424/changes#diff-58df0fd7dc8b2049e7daf12456ffd28d65ea5e66fbefa61151417744c2311902R52

.gitignore
mirkoCrobu: and where is Luigi?

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -115,12 +176,56 @@ func Load(platform platform.Platform, dir *paths.Path, modelsDir *paths.Path) (*
...
+	})
+	env := volumeEnv
+	for k, v := range envVars {
+		env = append(env, fmt.Sprintf("%s=%s", k, v))
+	}
+	err := dockerhandler.Run(ctx, cli, dockerhandler.RunOptions{
+		Image: handler.Image,
+		Cmd:   handler.Actions.Check,
+		Binds: binds,
+		Env:   env,
```
dido18: we need to parse the output to get the internal error and show them to the user

.gitignore
giulio93: to drop!!

internal/dockerhandler/dockerhandler.go
```diff
@@ -0,0 +1,112 @@
+// This file is part of arduino-app-cli.
+//
+// SPDX-FileCopyrightText: Arduino s.r.l. and/or its affiliated companies
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+// Package dockerhandler provides a thin Docker API wrapper for running a container
+// to completion and streaming its output.
+package dockerhandler
```
mirkoCrobu: file name should be docker_handler.go

internal/orchestrator/models.go
```diff
@@ -42,54 +43,51 @@ type AIModelItem struct {
 	Bricks            []string          `json:"brick_ids"`
 	Metadata          map[string]string `json:"metadata,omitempty"`
 	IsBuiltin         bool              `json:"is_builtin"`
-	DiskUsage         *uint64           `json:"disk_usage,omitempty"`
```
mirkoCrobu, line 45: We don't send this field anymore in the json response? Is this a breaking change for App Lab?

internal/api/handlers/ai_models.go
```diff
@@ -164,3 +168,58 @@ func (r InstallEIModelRequest) Validate() error {
 	}
 	return nil
 }
+
+func HandleInstallModel(dockerClient command.Cli, cfg config.Configuration, modelsIndex *modelsindex.ModelsIndex, plat platform.Platform) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		id := strings.TrimSpace(r.PathValue("modelID"))
```
lucarin91, line 174: Maybe just `id` is sufficient
```suggestion
		id := strings.TrimSpace(r.PathValue("id"))
```

internal/api/handlers/ai_models.go
```diff
@@ -164,3 +168,58 @@ func (r InstallEIModelRequest) Validate() error {
...
+		if id == "" {
+			render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "modelID must be set"})
```
lucarin91:
```suggestion
			render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "model ID must be set"})
```

internal/api/handlers/ai_models.go
```diff
@@ -164,3 +168,58 @@ func (r InstallEIModelRequest) Validate() error {
...
+		defer sseStream.Close()
+
+		type progress struct {
+			Name     string  `json:"name"`
+			Progress float32 `json:"progress"`
+		}
+		type log struct {
+			Message string `json:"message"`
+		}
+		if err := orchestrator.InstallModelByHandler(r.Context(), dockerClient.Client(), id, modelsIndex, cfg, plat, func(item orchestrator.StreamMessage) {
```
lucarin91: what byhandler means?
```suggestion
		if err := orchestrator.InstallModel(r.Context(), dockerClient.Client(), id, modelsIndex, cfg, plat, func(item orchestrator.StreamMessage) {
```

internal/orchestrator/handler_install.go
```diff
@@ -0,0 +1,95 @@
...
+		Env:   env,
+		LineCallback: func(line string) {
+			slog.Debug("handler output", "model", model.ID, "line", line)
+			parseHandlerLine([]byte(line), publish)
+		},
+		Stderr: io.Discard,
+	})
+}
+
+func parseHandlerLine(line []byte, publish func(ModelInstallEvent)) {
```
mirkoCrobu: it is identical with parseDownloadHandlerLine

internal/api/handlers/ai_models.go
mirkoCrobu: In line 180 you call `m, found, err := modelsIndex.GetModelByID(r.Context(), id)` 
you can pass `m `to `InstallModelByHandler` avoiding to call `GetModelByID `again inside `InstallModelByHandler`

internal/api/handlers/ai_models.go
```diff
@@ -164,3 +172,99 @@ func (r InstallEIModelRequest) Validate() error {
...
+			render.EncodeResponse(w, http.StatusConflict, models.ErrorResponse{Details: fmt.Sprintf("model %q already installed", id)})
+			return
+		}
+		if err := hasSufficientDiskSpace(cfg.AppsDir().Parent(), m.Size); err != nil {
+			if errors.Is(err, ErrInsufficientStorage) {
+				render.EncodeResponse(w, http.StatusInsufficientStorage, models.ErrorResponse{Details: "insufficient disk space"})
+				return
+			}
+			slog.Warn("unable to check disk space", slog.String("error", err.Error()))
+		}
```
lucarin91: I would add this check in the install `RunDownloadAction`, because this should be done also via the cli command

internal/api/handlers/ai_models.go
```diff
@@ -164,3 +172,99 @@ func (r InstallEIModelRequest) Validate() error {
...
+					sseStream.Send(render.SSEEvent{Type: "progress", Data: &progress{Name: m.ID, Progress: float32(e.Current) * 100 / float32(e.Total)}})
+				} else {
+					sseStream.Send(render.SSEEvent{Type: "progress", Data: &progress{Name: m.ID, Progress: 0}})
+				}
+			case modelsindex.ModelInstallEventComplete:
+				sseStream.Send(render.SSEEvent{Type: "progress", Data: &progress{Name: m.ID, Progress: 100}}) // used by the FE to understand that the installation is complete
+			case modelsindex.ModelInstallEventError:
+				sseStream.Send(render.SSEEvent{Type: "message", Data: log{Message: e.Description}})
+			default:
+				sseStream.Send(render.SSEEvent{Type: "message", Data: log{Message: e.Description}})
```
lucarin91: Why do we have an unmapped event? I would panic or return an error instead

```suggestion
			default:
			     panic("unkown event")
```

internal/api/handlers/ai_models.go
```diff
@@ -164,3 +172,99 @@ func (r InstallEIModelRequest) Validate() error {
...
+		}
+
+		err = modelsindex.RunDownloadAction(r.Context(), dockerClient.Client(), *m, handler, cfg, plat, installResponse)
+		if err != nil {
+			slog.Error("unable to install model", slog.String("error", err.Error()))
+			render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "unable to install model: " + err.Error()})
+			return
+		}
+	}
+}
```
lucarin91: I think we have too many events here; we can just start, progress instead, and handle completion and errors with the function return values.

```suggestion
		if err := modelsindex.RunDownloadAction(r.Context(), dockerClient.Client(), *m, handler, cfg, plat, func(e modelsindex.ModelDownloadEvent) {
			switch e.Type {
			case modelsindex.ModelInstallEventStart:
				sseStream.Send(render.SSEEvent{Type: "message", Data: log{Message: e.Description}})
			case modelsindex.ModelInstallEventUpdate:
				if e.Total > 0 {
					sseStream.Send(render.SSEEvent{Type: "progress", Data: &progress{Name: m.ID, Progress: float32(e.Current) * 100 / float32(e.Total)}})
				} else {
					sseStream.Send(render.SSEEvent{Type: "progress", Data: &progress{Name: m.ID, Progress: 0}})
				}
			default:
			     panic("unkown event")
 			}
		}); err != nil {
             sseStream.Send(render.SSEEvent{Type: "error", Data: err.Error()}})
			slog.Error("unable to install model", slog.String("error", err.Error()))
			render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "unable to install model: " + err.Error()})
			return
		} else {
		    sseStream.Send(render.SSEEvent{Type: "progress", Data: &progress{Name: m.ID, Progress: 100}}) // used by the FE to understand that the installation is complete
	}
}
```

internal/dockerhandler/dockerhandler.go
lucarin91: AFAIK, module names shouldn't have `_` or `-`

internal/orchestrator/model_install_manager.go
```diff
@@ -0,0 +1,30 @@
...
+
+const (
+	// Container-emitted event types (match the Python script's event field).
+	ModelInstallEventStart    ModelInstallEventType = "start"
+	ModelInstallEventUpdate   ModelInstallEventType = "update"
+	ModelInstallEventComplete ModelInstallEventType = "complete"
+	ModelInstallEventInfo     ModelInstallEventType = "info"
+	ModelInstallEventError    ModelInstallEventType = "error"
+	// Synthetic event emitted by Go after the container exits.
+	ModelInstallEventDone ModelInstallEventType = "done"
```
mirkoCrobu: we have the same consts in modelsindex/handlers_index.go?

internal/orchestrator/model_install_manager.go
```diff
@@ -0,0 +1,30 @@
...
+type ModelInstallEvent struct {
+	ModelID     string                `json:"model_id"`
+	Type        ModelInstallEventType `json:"type"`
+	Description string                `json:"description,omitempty"`
+	Current     int64                 `json:"current,omitempty"`
+	Total       int64                 `json:"total,omitempty"`
+	Unit        string                `json:"unit,omitempty"`
+	Percentage  string                `json:"percentage,omitempty"`
+	Artifacts   []string              `json:"artifacts,omitempty"`
+}
```
mirkoCrobu: same here: is it the same of `ModelDownloadEvent ` in modelsindex/handlers_index.go?

internal/dockerhelper/dockerhelper.go
lucarin91, line 1: Maybe this should be called `dockerhelper`?

internal/dockerhandler/dockerhandler.go
```diff
@@ -0,0 +1,112 @@
...
+			break
+		}
+		line := string(bytes.TrimSpace(w.buf[:idx]))
+		w.buf = w.buf[idx+1:]
+		if line != "" {
+			w.callback(line)
+		}
+	}
+	return len(b), nil
+}
```
lucarin91: you can use this instead https://github.com/arduino/arduino-app-cli/blob/85ccc639173e99efcd57de1db71926bb3a9f7dfd/internal/orchestrator/writer.go#L19, or import this better version from here https://github.com/bugst/go-algorithms/blob/master/io.go#L25

internal/dockerhandler/dockerhandler.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+type RunOptions struct {
+	Image        string
+	Cmd          []string
+	Binds        []string
+	Env          []string
+	Stdout       io.Writer
+	Stderr       io.Writer
+	LineCallback func(string)
```
lucarin91: I would remove the `LineCallback` here and make the caller use a CallbackWriter if they need it, like here https://github.com/arduino/arduino-app-cli/blob/e4add05db76435ff1f4fe4e1d1237c7ab0bc48ad/internal/update/apt/service.go#L185-L192
```suggestion
```
The current implementation is a missleading because the caller cannot pass both the lineCallback and the Stdout writer

internal/dockerhelper/dockerhandler.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	resp, err := cli.ContainerCreate(ctx,
+		&container.Config{
+			Image: opts.Image,
+			Cmd:   opts.Cmd,
+			Env:   opts.Env,
+		},
+		&container.HostConfig{
+			Binds: opts.Binds,
```
lucarin91: You could use the autoremove option to remove the container on exit
```suggestion
			Binds: opts.Binds,
			AutoRemove: true,
```

internal/dockerhandler/dockerhandler.go
```diff
@@ -0,0 +1,112 @@
...
+			Binds: opts.Binds,
+		},
+		nil, nil, "",
+	)
+
+	slog.Debug("creating container", "id", resp.ID, "image", opts.Image, "cmd", opts.Cmd, "env", opts.Env, "binds", opts.Binds)
+
+	if err != nil {
+		return fmt.Errorf("container create: %w", err)
+	}
```
lucarin91: You shouldn't read resp without checking the error, it could be an uninitialized pointer
```suggestion
	if err != nil {
		return fmt.Errorf("container create: %w", err)
	}
	slog.Debug("creating container", "id", resp.ID, "image", opts.Image, "cmd", opts.Cmd, "env", opts.Env, "binds", opts.Binds)
```

internal/dockerhandler/dockerhandler.go
```diff
@@ -0,0 +1,112 @@
...
+	defer cli.ContainerRemove(context.Background(), resp.ID, container.RemoveOptions{Force: true}) //nolint:errcheck
```
lucarin91:
```suggestion
```

internal/dockerhelper/dockerhandler.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	if opts.Stderr == nil {
+		opts.Stderr = io.Discard
+	}
+
+	resp, err := cli.ContainerCreate(ctx,
+		&container.Config{
+			Image: opts.Image,
+			Cmd:   opts.Cmd,
+			Env:   opts.Env,
```
lucarin91: To avoid permission issues, we should add the user and group that will run the container using this function, which we could move here https://github.com/arduino/arduino-app-cli/blob/52f38d8aaf5fadff8788ab57fd30e08c30d2d3d1/internal/orchestrator/orchestrator.go#L952-L963

```suggestion
            Env:    opts.Env,
			User: getCurrentUser(),
```

internal/dockerhelper/dockerhandler.go
```diff
@@ -0,0 +1,112 @@
...
+		return fmt.Errorf("reading output: %w", err)
+	}
+
+	inspect, err := cli.ContainerInspect(context.Background(), resp.ID)
+	if err != nil {
+		return fmt.Errorf("container inspect: %w", err)
+	}
+	if inspect.State.ExitCode != 0 {
+		return fmt.Errorf("container exited with status %d", inspect.State.ExitCode)
+	}
```
lucarin91: Instead of using the log driver, I will use attach and wait instead




```suggestion
	// Register the wait before starting to avoid missing the exit event.
	statusCh, errCh := cli.ContainerWait(ctx, resp.ID, container.WaitConditionNotRunning)

	attachResp, err := cli.ContainerAttach(ctx, resp.ID, container.AttachOptions{
		Stream: true,
		Stdout: true,
		Stderr: true,
	})
	if err != nil {
		return fmt.Errorf("container attach: %w", err)
	}
	defer attachResp.Close()

	if err := cli.ContainerStart(ctx, resp.ID, container.StartOptions{}); err != nil {
		return fmt.Errorf("container start: %w", err)
	}

	if _, err := stdcopy.StdCopy(opts.Stdout, opts.Stderr, attachResp.Reader); err != nil {
		return fmt.Errorf("reading output: %w", err)
	}

	select {
	case err := <-errCh:
		if err != nil {
			return fmt.Errorf("container wait: %w", err)
		}
	case status := <-statusCh:
		if status.Error != nil {
			return fmt.Errorf("container exit error: %s", status.Error.Message)
		}
		if status.StatusCode != 0 {
			return fmt.Errorf("container exited with status %d", status.StatusCode)
		}
	}

```

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -59,53 +88,94 @@ type BrickConfig struct {
...
-func (m *ModelsIndex) GetModels() []AIModel {
-	return m.loadModels()
+func (m *ModelsIndex) GetModelsByBricks(ctx context.Context, bricks []string) []AIModel {
+	return m.filterByBricks(bricks)
 }
 
-func (m *ModelsIndex) GetModelByID(id string) (*AIModel, bool) {
-	models := m.loadModels()
+// FindModelByID returns the model with the given ID without checking installed status.
+func (m *ModelsIndex) FindModelByID(id string) (*AIModel, bool) {
```
lucarin91: not sure if we should keep this function, it seems confusing

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -48,7 +74,10 @@ type AIModel struct {
 	ModelLabels       []string          `yaml:"model_labels,omitempty"`
 	Metadata          map[string]string `yaml:"metadata,omitempty"`
 	IsInternal        bool              `yaml:"-"`
+	Installed         bool              `yaml:"-"`
+	Size              uint64            `yaml:"-"`
```
lucarin91: I would instead add an `Info` method for getting both `Installed` and `Size` information

```go
type AIModelInfo struct {
    Installed         bool  
    Size              uint64 
}

func (m *AIModel) Info() (AIModelInfo, error) {
    // ....
}
```

internal/api/handlers/ai_models.go
dido18: this shoul be resovled in https://github.com/arduino/arduino-app-cli/pull/457 already merged

internal/orchestrator/modelsindex/handlers_index.go
```diff
@@ -0,0 +1,350 @@
...
+	bindPath := downloadPath
+	if modelRepo != "" {
+		bindPath = downloadPath.Join(modelRepo)
+	}
+	if err := bindPath.MkdirAll(); err != nil {
+		return fmt.Errorf("cannot create model directory %q: %w", bindPath, err)
+	}
+	if err := os.Chmod(bindPath.String(), 0777); err != nil { //nolint:gosec
+		slog.Warn("cannot set permissions on model directory", "path", bindPath, "err", err)
+	}
```
lucarin91: this should be necessary

```suggestion
```

lucarin91 (commented):
I would provide some initial comments, nothing major, but I think it would keep things more consistent. I would then need to do an in-depth review of the BrickIndex module and the new handler.

internal/orchestrator/models.go
giulio93: We are using Size now, i don't remember why we call it DiskUsage in the first place.

internal/orchestrator/handler_install.go
giulio93: What do you propose to unify them?

internal/orchestrator/model_install_manager.go
giulio93: what do you propose to unify them?

internal/dockerhelper/dockerhandler.go
giulio93: With the `Follow` option the driver logs stream the logs from the start even if the container is already exited.
Since the container not always has a long run, it could be better to use the log driver

internal/api/handlers/ai_models.go
giulio93: All the endpoint has v1/models/{modelID} do we want to change that?

lucarin91: Isn't the model redundant there?

internal/dockerhelper/dockerhandler.go
lucarin91: The opposite, by using attach, we are faster because the stdout and stderr of the container don't go inside the log driver and are directly output in the stdout and err of the process. Actually, we should make sure we clean up all logs of this container, which aren't of much use.

internal/dockerhelper/dockerhandler.go
```diff
@@ -0,0 +1,147 @@
...
+
+	resp, err := cli.ContainerCreate(ctx,
+		&container.Config{
+			Image: opts.Image,
+			Cmd:   opts.Cmd,
+			Env:   opts.Env,
+			User:  getCurrentUser(),
+		},
+		&container.HostConfig{
+			Binds: opts.Binds,
```
lucarin91: I would also disable logs, to avoid waisting space
```suggestion
			Binds: opts.Binds,
			LogConfig: container.LogConfig{Type: "none"},
```

internal/dockerhandler/dockerhandler.go
giulio93: imported

internal/orchestrator/model_install_manager.go
lucarin91: You can just remove this file, those types are never used

internal/orchestrator/modelsindex/handlers_index.go
```diff
@@ -0,0 +1,489 @@
...
+type ModelInstallEventType string
+
+const (
+	ModelInstallEventStart    ModelInstallEventType = "start"
+	ModelInstallEventUpdate   ModelInstallEventType = "update"
+	ModelInstallEventComplete ModelInstallEventType = "complete"
+	ModelInstallEventInfo     ModelInstallEventType = "info"
+	ModelInstallEventError    ModelInstallEventType = "error"
+	ModelInstallEventDone     ModelInstallEventType = "done"
+)
```
lucarin91: this is never used and can be removed
```suggestion
```

internal/orchestrator/modelsindex/handlers_index.go
```diff
@@ -0,0 +1,489 @@
...
+	DiskSizeMB  *float64 `json:"disk_size_mb"`  // actual on-disk size, only when installed
+}
+
+var (
+	ErrNotFound            = errors.New("model not found")
+	ErrConflict            = errors.New("can't delete the model")
+	ErrCannotRemoveModel   = errors.New("cannot remove an internal model")
+	ErrInsufficientStorage = errors.New("insufficient storage to install the model")
+	ErrIncompleteImpulse   = errors.New("impulse not ready for deployment")
+)
```
lucarin91: those are never used and can be removed
```suggestion
```

internal/orchestrator/modelsindex/handlers_index.go
```diff
@@ -0,0 +1,489 @@
...
+		}
+	}
+	return modelsInfo, nil
+}
+
+func runInfoAction(ctx context.Context, cli client.APIClient, handler ModelHandler, model AIModel, plat platform.Platform, configEnv map[string]string) (uint64, error) {
+	var env []string
+	if plat.BoardName != "" {
+		env = append(env, fmt.Sprintf("board=%s", plat.BoardName))
+	}
```
lucarin91: This can be removed also because the env variable should be BOARD_NAME afaik
```suggestion
```

internal/orchestrator/modelsindex/handlers_index.go
```diff
@@ -0,0 +1,489 @@
...
+		Binds: ResolveVarsSlice(binds, configEnv),
+		Env:   env,
+		Stdout: f.NewCallbackWriter(func(line string) {
+			var out struct {
+				Event  string  `json:"event"`
+				SizeMB float64 `json:"size_mb"`
+			}
+			if jsonErr := json.Unmarshal([]byte(line), &out); jsonErr == nil && out.Event == "stat" && out.SizeMB > 0 {
+				size = uint64(out.SizeMB * 1024 * 1024)
+			}
```
lucarin91, line 239: We should have two utility functions in `dockerhelper`, one for handling streaming and the other for handling a single command. I would propose a PR about that

internal/e2e/daemon/ai_models_handler_test.go
```diff
@@ -0,0 +1,160 @@
+// This file is part of arduino-app-cli.
+//
+// SPDX-FileCopyrightText: Arduino s.r.l. and/or its affiliated companies
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+//go:build e2e_docker
+
```
lucarin91: I would make this test run
```suggestion
```

internal/e2e/daemon/ai_models_handler_test.go
```diff
@@ -0,0 +1,160 @@
...
+//   - SSE progress events (start / update / complete / done) are emitted and parsed
+//   - The model is reported as installed after download
+//   - The delete container runs without "short write" errors
+//
+// Requires E2E_MODEL_ID env var (e.g. "melo-tts-es") and a running Docker daemon.
+func TestModelHandlerDownloadFlow(t *testing.T) {
+	modelID := os.Getenv("E2E_MODEL_ID")
+	if modelID == "" {
+		t.Skip("set E2E_MODEL_ID to run Docker-backed model handler tests")
+	}
```
lucarin91: Not sure why we want to pass the model id from env I would add a single model or a list of model to test instead
```suggestion
	modelID := "melo-tts-es"
```

internal/api/api.go
```diff
@@ -62,9 +62,10 @@ func NewHTTPRouter(
 	mux.Handle("PUT /v1/system/update/apply", handlers.HandleUpdateApply(updater))
 	mux.Handle("GET /v1/system/resources", handlers.HandleSystemResources(cfg))
 
-	mux.Handle("GET /v1/models", handlers.HandleModelsList(modelsIndex))
-	mux.Handle("GET /v1/models/{modelID}", handlers.HandlerModelByID(modelsIndex))
+	mux.Handle("GET /v1/models", handlers.HandleModelsList(dockerClient, modelsIndex, cfg, platform))
+	mux.Handle("GET /v1/models/{modelID}", handlers.HandlerModelByID(dockerClient, modelsIndex, cfg, platform))
 	mux.Handle("PUT /v1/models/ei/projects/{projectID}", handlers.HandleInstallEIModel(cfg, bricksIndex, modelsIndex, dockerClient))
+	mux.Handle("PUT /v1/models/{modelID}", handlers.HandleInstallModel(dockerClient, cfg, modelsIndex, platform))
```
dido18: Nit 

use the same order of the other handlers
```suggestion
	mux.Handle("PUT /v1/models/{modelID}", handlers.HandleInstallModel(dockerClient, modelsIndex, cfg, platform))
```

internal/orchestrator/bricks/bricks.go
```diff
@@ -88,7 +89,7 @@ func (s *Service) AppBrickInstancesList(a *app.ArduinoApp, platform platform.Pla
 			ModelID:         cmp.Or(brickInstance.Model, brick.GetModelNameByBoard(platform.BoardName)),
 			Variables:       variablesMap,
 			ConfigVariables: configVariables,
-			CompatibleModels: f.Map(s.modelsIndex.GetModelsByBrick(brick.ID), func(m modelsindex.AIModel) AIModel {
+			CompatibleModels: f.Map(s.modelsIndex.GetModelsByBrick(context.TODO(), brick.ID), func(m modelsindex.AIModel) AIModel {
```
dido18: we could avoid spinnnig a container fo the brick thas has `RequireModel` to false.

```
func (b Brick) GetCompatibleModels(contex.ctx, modelsindex.AIModel) []AiModel{
   var models []AiModel
   if b.RequireModel == false{
     return models
  }

...
}


```

dido18: resolved in perr coding

internal/api/handlers/ai_models.go
```diff
@@ -164,3 +172,93 @@ func (r InstallEIModelRequest) Validate() error {
...
+		}
+		if model == nil {
+			render.EncodeResponse(w, http.StatusNotFound, models.ErrorResponse{Details: fmt.Sprintf("model %q not found", id)})
+			return
+		}
+		if model.Installed {
+			render.EncodeResponse(w, http.StatusConflict, models.ErrorResponse{Details: fmt.Sprintf("model %q already installed", id)})
+			return
+		}
+		if err := hasSufficientDiskSpace(cfg.AppsDir().Parent(), model.Size); err != nil {
```
dido18: we should move this check inside the ` modelsIndex.Download` to have it also in the cli command

internal/api/handlers/ai_models.go
```diff
@@ -31,29 +35,33 @@ type InstallEIModelRequest struct {
 	ImpulseID *int `json:"impulse_id" description:"Edge Impulse impulse ID" example:"1" required:"true"`
 }
 
-func HandleModelsList(modelsIndex *modelsindex.ModelsIndex) http.HandlerFunc {
+func HandleModelsList(dockerClient command.Cli, modelsIndex *modelsindex.ModelsIndex, cfg config.Configuration, plat platform.Platform) http.HandlerFunc {
```
dido18, line 34:
```suggestion
func HandleModelsList(dockerClient client.APIClient, modelsIndex *modelsindex.ModelsIndex, cfg config.Configuration, plat platform.Platform) http.HandlerFunc {
```

internal/api/handlers/ai_models.go
```diff
@@ -31,29 +35,33 @@ type InstallEIModelRequest struct {
...
 	return func(w http.ResponseWriter, r *http.Request) {
 		params := r.URL.Query()
 
 		var brickFilter []string
 		if brick := params.Get("bricks"); brick != "" {
 			brickFilter = strings.Split(strings.TrimSpace(brick), ",")
 		}
-		res := orchestrator.AIModelsList(orchestrator.AIModelsListRequest{
+		res := orchestrator.AIModelsList(r.Context(), dockerClient.Client(), orchestrator.AIModelsListRequest{
```
dido18, line 42:
```suggestion
		res := orchestrator.AIModelsList(r.Context(), dockerClient, orchestrator.AIModelsListRequest{
```

dido18 (commented):
The e2e tests runs only on `arm64` becasue the endpoints to download the models uses the `ghcr.io/arduino/app-bricks/models-downloader:` that is built for arm64 archtecture.
```
WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64/v3) and no specific platform was requested
exec /app/ai_hub/ai_hub_model_checker.sh: exec format erro
```

we coul skip the test  
```
if runtime.GOARCH != "arm64" {
		t.Skipf("Skipping test: requires arm64 architecture, currently running on %s", runtime.GOARCH)
	}

File: debian/arduino-app-cli/etc/systemd/system/arduino-app-cli.service [hunk 44]
```diff
@@ -4,7 +4,7 @@ After=network-online.target docker.service arduino-router.service
 Wants=network-online.target docker.service arduino-router.service
 
 [Service]
-ExecStart=/usr/bin/arduino-app-cli daemon --port 8800 --log-level error
+ExecStart=/usr/bin/arduino-app-cli daemon --port 8800 --log-level debug
```
dido18:
```suggestion
ExecStart=/usr/bin/arduino-app-cli daemon --port 8800 --log-level error
```

internal/api/docs/openapi.yaml
```diff
@@ -1449,12 +1450,10 @@ components:
           type: array
         description:
           type: string
-        disk_usage:
```
dido18, line 1452: check that is is not a breaking change

internal/e2e/e2e.go
```diff
@@ -47,12 +46,19 @@ func WithCustomModelDir(dir *paths.Path) ArduinoAppCLIOption {
...
 }
 
+func WithBoardName(name string) ArduinoAppCLIOption {
+	return func(cli *ArduinoAppCLI) {
+		dataDir := paths.New(cli.envVars["ARDUINO_APP_CLI__DATA_DIR"])
+		content := fmt.Sprintf(`{"board_name":%q}`, name)
+		cli.t.NoError(dataDir.Join("platform.json").WriteFile([]byte(content)))
+	}
+}
+
```
dido18, line 56:
```suggestion
func WithBoardName(cfg config. Configuration, name string) ArduinoAppCLIOption {
	return func(cli *ArduinoAppCLI) {
		content := fmt.Sprintf(`{"board_name":%q}`, name)
		dataDir := cfg.DataDir()
		cli.t.NoError(dataDir.Join("platform.json").WriteFile([]byte(content)))
	}
}

```

Taskfile.yml
```diff
@@ -197,6 +197,24 @@ tasks:
...
+      APP_PORT: '{{.APP_PORT | default "8801"}}'
+    env:
+      GOTOOLCHAIN: local
+    cmds:
+      - CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -gcflags="all=-N -l" -o {{.DEBUG_BINARY}} ./cmd/arduino-app-cli
+      - adb push {{.DEBUG_BINARY}} {{.BOARD_PATH}}
+      - adb shell chmod +x {{.BOARD_PATH}}
+      - adb forward tcp:{{.DLV_PORT}} tcp:{{.DLV_PORT}}
+      - adb forward tcp:{{.APP_PORT}} tcp:{{.APP_PORT}}
+      - adb shell "/home/arduino/go/bin/dlv --listen=:{{.DLV_PORT}} --headless --api-version=2 exec {{.BOARD_PATH}} -- daemon --port {{.APP_PORT}} --log-level debug"
```
dido18: Is this useful and working ? otherwise we could remove it

internal/orchestrator/bricks/testdata/dummy-app-brick-override-temp/app.yaml
```diff
@@ -1,4 +1,9 @@
 name: Copy of Blinking LED from Arduino Cloud
 description: Control the LED from the Arduino IoT Cloud using RPC calls
 ports: []
+bricks:
+- arduino:arduino_cloud:
+    variables:
+      ARDUINO_DEVICE_ID: this-is-a-device-id
+      ARDUINO_SECRET: this-is-a-secret
```
lucarin91: this should be reverted

Taskfile.yml
giulio93: removed

internal/api/docs/openapi.yaml
giulio93: DiskUsage https://github.com/arduino/arduino-app-cli/blob/main/internal/orchestrator/models.go#L103
It was calculated visiting the model directory. Now the size is present in the model-list.yaml, if not it's calculated calling the info container action.
DiskUsage has been substitute by Size.

lucarin91 (approved):

PR #433, opened 2026-06-04
Add NPU usage to the system/resources API (embedded lib)

martacarbone:
### Motivation

Use the libnpuperf to return to the App Lab some stats about the usage of the NPU.
This PR use purego to import the C API that is in chage to get the metrics from the DSP.
Go use this library allowing a shared access to the metric api by ensure a single request at time and keeping the allocated resources as soon there are active streams, by a reference counter.
The api calls should be ties to a single OS thread.

The stream works if the npu is not available (running on unoq) and if the fastrpc protocol fails for some reasons.

### Reviewer checklist


internal/orchestrator/resources/npu.go
```diff
@@ -0,0 +1,176 @@
...
+type npuRequest struct {
+	responseType npuRequestType
+	err          chan error
+	maxUsage     chan float32
+}
+
+var (
+	npuWorkerOnce sync.Once
+	npuRequests   chan npuRequest
+	refCounter    int // no need atomic here
```
martacarbone: This is used by a single call in the same thread, so no need of atomic here. This is used to handle multiple stream connected, avoiding to free resources while there are still active connected streams.
We can get rid of this by deallocating at system (arduino-all-cli) shutdown.

internal/orchestrator/resources/npu.go
```diff
@@ -0,0 +1,176 @@
...
+	npuWorkerOnce.Do(func() {
+		npuRequests = make(chan npuRequest)
+		go npuWorker()
+	})
+}
+
+// npuWorker is the dedicated goroutine that locks to a single OS thread.
+// It processes all NPU requests sequentially, ensuring thread-local state consistency.
+func npuWorker() {
+	runtime.LockOSThread()
```
martacarbone, line 57: This is required because the C func we are calling uses (among the other calls) remote_handle64_open, see [here](https://github.com/qualcomm/libqcnpuperf/blob/main/src/qcom_dsp.c#L54)  and this is no thread safe, see [here](https://github.com/qualcomm/fastrpc/blob/development/inc/remote.h#L799)

internal/orchestrator/resources/npu.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	var noMetrics int32
+	ptr := qcomDspGetProfData(DSP_NPU0, &noMetrics)
+	if ptr == nil || noMetrics <= 0 {
+		return 0, fmt.Errorf("qcomDspGetProfData error: no profile data available")
+	}
+
+	data := (*SysmonQueryProfData)(ptr)
+	return max(data.Q6Utilization, data.HmxUtilization, data.HvxUtilization), nil
```
martacarbone, line 113: We can move this outside, letting the called to choice which is the best value to be used as output

internal/orchestrator/resources/npu.go
```diff
@@ -0,0 +1,176 @@
...
+
+var (
+	npuWorkerOnce sync.Once
+	npuRequests   chan npuRequest
+	refCounter    int // no need atomic here
+)
+
+func startNPUWorker() {
+	npuWorkerOnce.Do(func() {
+		npuRequests = make(chan npuRequest)
```
martacarbone, line 47: Allows only one call at time

internal/orchestrator/resources/libnpu.go
```diff
@@ -0,0 +1,76 @@
...
+
+// There are the three enpoints provided by libqcnpuperf.so
+var (
+	qcomDspInit        func(int32) int32
+	qcomDspGetProfData func(int32, *int32) unsafe.Pointer
+	qcomDspDeinit      func(int32) int32
+)
+
+var libqcnpuperfLib uintptr
+var libqcnpuperfLibraryPath = "/var/lib/arduino-app-cli/libqcnpuperf.so"
```
martacarbone, line 51: This is temporary here. We should:
1. understand if this can be provided by an external debian package and add it as ventunoQ dependency
2. generate ourselves the library and put it in the ventunoQ metapackage

martacarbone: Opted for solution #1, added to arduino-ventunoQ-config package

internal/orchestrator/resources/npu.go
martacarbone: Remove the reference counter. The struct is initialized the first time it is requested and remains available for the entire lifetime of the program.

martacarbone:
Closing this issue because we chose to use an external binary for the implementation.

PR #466, opened 2026-06-18
fix: follow link in app export

### LINKED ISSUE #448: export app fails if the app has simlink
### Describe the problem

Adding a link to a file makes the export fail

```
failed to create zip archive: read /home/arduino/ArduinoApps/particle-generator/python2: is a directory
```

### To reproduce

1. Inside an app writes `ln -s python python2`
2. export the app with arduino-app-cli app export my-app
3. The export will fail 

### Expected behavior

The zip of the app will have the linked folder or file duplicated, mimicking the behavior of the zip/unzip Linux utility.

### Arduino App CLI version

NAN

### Additional context

_No response_

### Issue checklist

lucarin91:
### Motivation

App export fails if the app folder has some links.

### Change description

Add the link following in the zip creation. 

Removed [filepath.WalkDir](https://pkg.go.dev/path/filepath#WalkDir), which doesn't follow links and uses https://github.com/arduino/go-paths-helper/blob/master/readdir.go#L78

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/archive.go
+++ b/internal/orchestrator/archive.go
@@ -12,7 +12,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


internal/orchestrator/archive_test.go
```diff
@@ -78,13 +79,34 @@ func TestExportAppZip(t *testing.T) {
...
+		{
+			name:         "Symlink to directory is included in zip",
+			appName:      "Symlink Dir App",
+			files:        []string{"app.yaml", "data/foo.txt"},
+			symlinks:     map[string]string{"data2": "data"},
+			includeData:  true,
+			wantErr:      false,
+			wantFilename: "symlink-dir-app.zip",
+			wantFiles:    []string{"app.yaml", "data/foo.txt", "data2/foo.txt"},
+		},
```
mirkoCrobu, line 101: we could test also the symlink loops
```suggestion
		},
		{
    name:    "Circular symlink does not loop",
    appName: "Circular App",
    files:   []string{"app.yaml"},
    symlinks: map[string]string{
        "link-a": "link-b",
        "link-b": "link-a",
    },
    includeData:  false,
    wantErr:      true, 
    wantFilename: "circular-app.zip",
    wantFiles:    []string{"app.yaml"},
},
```

internal/orchestrator/archive.go
```diff
@@ -52,37 +51,40 @@ func zipAppToBuffer(bricksIndex *bricksindex.BricksIndex, sourcePath string, roo
...
-			if !includeData && name == "data" {
-				return filepath.SkipDir
-			}
+			zipWriter.Close()
+			return nil, err
 		}
 
-		info, err := d.Info()
+		info, err := entry.Stat() // follows symlinks
 		if err != nil {
```
mirkoCrobu, line 80: if we have symlinks, the error seems to be: 
`failed to create zip archive: stat /tmp/.../link-a: too many levels of symbolic links`
Should we return a clearer error like the following? 


```suggestion
if err != nil {
    if errors.Is(err, syscall.ELOOP) {
        return nil, fmt.Errorf("circular symlink detected: %s", entry.String())
    }
    zipWriter.Close()
    return nil, err
}
```

lucarin91: Thanks for the report. I would actually skip the circular links. I think it is better to try exporting an app anyway instead of failing.

mirkoCrobu (approved):

internal/orchestrator/archive.go
mirkoCrobu: I agree!

PR #473, opened 2026-06-23
feat(logs): return better parsable fields and service support

lucarin91:
### Motivation

The `id` field in `/v1/apps/{id}/logs` currently exposes the underlying container reference, which has no meaning in our system. It should be one of our own identifiers so consumers can tell whether a log line comes from the main python container or a brick.

### Change description

- `id` = `"main"` for the app's python container, or the brick ID for brick containers.
- New `container_id` field carries the underlying container reference (kept for debugging).
- Filter rename: `app` → `main`, `services` → `bricks`.

**Breaking:** filter values and response fields. app-lab does not use the `filter` param, so it should not see any meaningful change in the response.

### Reviewer checklist


internal/api/handlers/app_logs.go
```diff
@@ -42,10 +52,10 @@ func HandleAppLogs(
 
 		queryParams := r.URL.Query()
 
-		showAppLogs, showServicesLogs := true, false
+		showAppLogs, showServicesLogs := true, true
 		if filter := queryParams.Get("filter"); filter != "" {
 			filters := strings.Split(strings.TrimSpace(filter), ",")
-			showServicesLogs = slices.Contains(filters, "services")
+			showServicesLogs = slices.Contains(filters, "bricks")
```
dido18, line 56: the bricks python code is executed inside the main container, the bricks python logic is shon if `showAppLogs=true`

The filter by "bricks" does not what I expecte, it only filter the services of the bricks

Xayton (approved):

PR #476, opened 2026-06-23
Add arduino-app-cloud socket dir

fabik111:
### Motivation
Add the socket for communication between App and the Arduino-app-cloud
### Reviewer checklist

### DIFF
--- a/internal/orchestrator/config/config.go
+++ b/internal/orchestrator/config/config.go
@@ -26,6 +26,7 @@ type Configuration struct {
 	appsDir                          *paths.Path
 	dataDir                          *paths.Path
 	routerSocketPath                 *paths.Path
+	cloudDaemonSocketDir             *paths.Path
 	customModelsDir                  *paths.Path
 	PythonImage                      string
 	UsedPythonImageTag               string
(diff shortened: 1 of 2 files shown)


internal/orchestrator/config/config.go
```diff
@@ -64,6 +65,14 @@ func NewFromEnv() (Configuration, error) {
 		routerSocket = paths.New("/var/run/arduino-router.sock")
 	}
 
+	// Directory holding the arduino-app-cloud daemon's UNIX socket; bind-mounted
+	// (the directory, so it stays valid across daemon restarts) into app
+	// containers so the arduino_cloud brick can reach the daemon.
+	cloudDaemonSocketDir := paths.New(os.Getenv("ARDUINO_CLOUD_DAEMON_SOCKET_DIR"))
```
mirkoCrobu, line 71: Here the env vars follow mixed conventions (ARDUINO_APP_CLI__*, ARDUINO_APP_BRICKS__*, ARDUINO_ROUTER_SOCKET, plus unprefixed ones).

Should we standardize on a single namespace? For this socket could be:
`ARDUINO_APP_CLI__CLOUD_DAEMON_SOCKET_DIR`, or
`ARDUINO_APP_BRICKS__CLOUD_DAEMON_SOCKET_DIR `

We could align this one now and rename the others in a dedicated follow-up PR.

lucarin91: Yeah, that is my biggest concern about this PR. If it isn't particularly required, I would just hardcode the value without adding an env variable

mirkoCrobu: Good solution. In addition, do you think it is better to add a "TODO" comment, or should we open an issue (or both)?

fabik111: So, what's the final stand? Should I hardcode the path without using env var?

lucarin91: I think we converge on a new solution; I would do a PR for this.

lucarin91:
Superseded by https://github.com/arduino/arduino-app-cli/pull/490

PR #478, opened 2026-06-24
[CLI] adding inline json output stream for init cmd

mirkoCrobu:
### Motivation

closes https://github.com/bcmi-labs/orchestrator/issues/521 (check it for the full context, limitations, integration, realated issue) 

### Change description

- Single typed `InitEvent callback` (log/progress) as the only path to `stdout` for system init.
- Per-layer Docker logs removed and replaced by progress aggregated globally across all images.
- Added `arduino-app-cli system init --format json-lines` flag to system init(extending feedback pkg. this could be used for system update command too), emitting JSONL with `type`, `source`, `label`, `current`, `total`, `percent `fields.
- The `source ` field (docker/libs/platform) it will be used by the consumers (it will be used in the upgrade flow, where system init is called as a sub-process to distinguish between the update source events): I preferred to keep all the logs instead of returning just the Docker ones to avoid losing logs.
- Disk space check refactoring(GetBytesToDownload removed).

### Reviewer checklist

### DIFF
--- a/cmd/arduino-app-cli/system/system.go
+++ b/cmd/arduino-app-cli/system/system.go
@@ -50,10 +50,21 @@ func newDownloadImageCmd(cfg config.Configuration) *cobra.Command {
 		Args:   cobra.ExactArgs(0),
 		Hidden: true,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			return orchestrator.SystemInit(cmd.Context(), cfg, servicelocator.GetPlatform(), servicelocator.GetBricksIndex(), servicelocator.GetServicesIndex(), servicelocator.GetModelsIndex(), servicelocator.GetDockerClient(), orchestrator.SystemInitOptions{
+			stdout, _, err := feedback.DirectStreams()
+			if err != nil {
+				return err
+			}
+			progressCB := func(progress orchestrator.InitProgress) {
+				percentage := float64(progress.Curr) / float64(progress.Total) * 100
+				fmt.Fprintf(stdout, "%s: %.2f%% (%d/%d)\r", progress.Label, percentage, progress.Curr, progress.Total)
+				if progress.Curr == progress.Total {
+					fmt.Fprintln(stdout)
+				}
+			}
+			return orchestrator.SystemInit(cmd.Context(), cfg, servicelocator.GetPlatform(), servicelocator.GetBricksIndex(), servicelocator.GetServicesIndex(), servicelocator.GetDockerClient(), servicelocator.GetModelsIndex(), orchestrator.SystemInitOptions{
 				OnlyDockerImages:    onlyImages,
 				OnlyPlatformAndLibs: onlyPlatformAndLibraries,
-			})
+			}, progressCB)
 		},
 	}
 
--- a/cmd/arduino-app-cli/system/system.go
+++ b/cmd/arduino-app-cli/system/system.go
@@ -6,9 +6,7 @@
 package system
 
 import (
-	"encoding/json"
 	"fmt"
-	"io"
 	"slices"
 	"strings"
 
@@ -44,89 +42,10 @@ func NewSystemCmd(cfg config.Configuration) *cobra.Command {
 	return cmd
 }
 
-// jsonInitEvent is the JSON-lines representation of a single event emitted by
-// `system init --json-feedback`. The Type field ("log" | "progress") is a
-// discriminator that tells the consumer which fields are meaningful.
-type jsonInitEvent struct {
-	Type    string `json:"type"`
-	Message string `json:"message,omitempty"`
-	Label   string `json:"label,omitempty"`
-	Current int64  `json:"current,omitempty"`
-	Total   int64  `json:"total,omitempty"`
-	Percent int    `json:"percent,omitempty"`
-}
-
-// newInitEventCallback builds the single callback that renders SystemInit events
-// to stdout. The same events are emitted regardless of the flag; only the
-// formatting differs (plain text lines vs JSON lines). Progress events are
-// throttled so consecutive updates with the same integer percentage are dropped.
-func newInitEventCallback(stdout io.Writer, jsonFeedback bool) orchestrator.InitEventCallback {
-	var render orchestrator.InitEventCallback
-	if jsonFeedback {
-		// One JSON object per line (JSONL). json.Encoder.Encode writes a trailing
-		// newline after each object, which is exactly the framing a line-based
-		// consumer needs.
-		enc := json.NewEncoder(stdout)
-		render = func(e orchestrator.InitEvent) {
-			switch e.Type {
-			case orchestrator.InitLogEvent:
-				_ = enc.Encode(jsonInitEvent{Type: "log", Message: e.Message})
-			case orchestrator.InitProgressEvent:
-				p := e.Progress
-				var percentage int
-				if p.Total > 0 {
-					percentage = int(float64(p.Curr) / float64(p.Total) * 100)
-				}
-				_ = enc.Encode(jsonInitEvent{
-					Type:    "progress",
-					Label:   p.Label,
-					Current: p.Curr,
-					Total:   p.Total,
-					Percent: percentage,
-				})
-			}
-		}
-	} else {
-		render = func(e orchestrator.InitEvent) {
-			switch e.Type {
-			case orchestrator.InitLogEvent:
-				fmt.Fprintln(stdout, e.Message)
-			case orchestrator.InitProgressEvent:
-				p := e.Progress
-				percentage := float64(p.Curr) / float64(p.Total) * 100
-				fmt.Fprintf(stdout, "%s: %.0f%%\n", p.Label, percentage)
-			}
-		}
-	}
-	return throttleProgress(render)
-}
-
-// throttleProgress wraps an event callback so that consecutive progress events
-// for the same label are only forwarded when their integer percentage changes.
-// Log events always pass through. This bounds the output volume (~100 progress
-// lines per label) without dropping any log line, identically in both formats.
-func throttleProgress(next orchestrator.InitEventCallback) orchestrator.InitEventCallback {
-	lastPct := map[string]int{}
-	return func(e orchestrator.InitEvent) {
-		if e.Type == orchestrator.InitProgressEvent {
-			p := e.Progress
-			if p.Total <= 0 {
-				return
-			}
-			pct := int(float64(p.Curr) / float64(p.Total) * 100)
-			if last, ok := lastPct[p.Label]; ok && last == pct {
-				return
-			}
-			lastPct[p.Label] = pct
-		}
-		next(e)
-	}
-}
-
 func newDownloadImageCmd(cfg config.Configuration) *cobra.Command {
 	var onlyImages bool
 	var onlyPlatformAndLibraries bool
-	var jsonFeedback bool
+	var jsonStream bool
 	cmd := &cobra.Command{
 		Use:    "init",
 		Args:   cobra.ExactArgs(0),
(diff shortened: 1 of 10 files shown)


mirkoCrobu:
### arduino-app-cli system init --format json-lines

image
image

mirkoCrobu:
###  arduino-app-cli system init
image

cmd/arduino-app-cli/system/system.go
```diff
@@ -45,20 +45,27 @@ func NewSystemCmd(cfg config.Configuration) *cobra.Command {
...
 				OnlyDockerImages:    onlyImages,
 				OnlyPlatformAndLibs: onlyPlatformAndLibraries,
-			})
+			}, eventCB)
 		},
 	}
 
 	cmd.PersistentFlags().BoolVar(&onlyImages, "only-docker-images", false, "Only download the application docker images")
 	cmd.PersistentFlags().BoolVar(&onlyPlatformAndLibraries, "only-arduino-platform", false, "Only download the Arduino platform and libraries")
+	cmd.PersistentFlags().BoolVar(&jsonStream, "json-stream", false, "Emit logs and progress as JSON lines (one event per line) instead of plain text")
```
lucarin91: I would use the format flag that we have in someother places, e.g., `--format json-lines`

cmd/arduino-app-cli/system/system_helper.go
```diff
@@ -0,0 +1,81 @@
...
+	Label   string `json:"label,omitempty"`
+	Current int64  `json:"current,omitempty"`
+	Total   int64  `json:"total,omitempty"`
+	Percent int    `json:"percent,omitempty"`
+}
+
+func newInitEventCallback(stdout io.Writer, jsonStream bool) orchestrator.InitEventCallback {
+	var render orchestrator.InitEventCallback
+	if jsonStream {
+		enc := json.NewEncoder(stdout)
```
lucarin91: What do you think about using the same approach of `PrintResult` https://github.com/arduino/arduino-app-cli/blob/0cb771e71849eb6040a0b6b3b924f8bd54fa71a6/cmd/feedback/feedback.go#L223? We can define a struct that implements the `Result` interface and use that here

cmd/arduino-app-cli/system/system.go
mirkoCrobu: ok I was not sure to use the same flag since it is a different output type, but "json-lines" instead of just "json" is perfect!

cmd/arduino-app-cli/system/system_helper.go
mirkoCrobu: done

cmd/arduino-app-cli/system/system.go
```diff
@@ -50,10 +50,16 @@ func newDownloadImageCmd(cfg config.Configuration) *cobra.Command {
 		Args:   cobra.ExactArgs(0),
 		Hidden: true,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			return orchestrator.SystemInit(cmd.Context(), cfg, servicelocator.GetPlatform(), servicelocator.GetBricksIndex(), servicelocator.GetServicesIndex(), servicelocator.GetModelsIndex(), servicelocator.GetDockerClient(), orchestrator.SystemInitOptions{
+			printEvent, err := feedback.NewResultStream()
+			if err != nil {
+				return err
+			}
+
+			eventCB := newInitEventCallback(printEvent)
```
lucarin91: I would rewrite this as follows
```
return orchestrator.SystemInit(
	cmd.Context(),
	cfg,
	servicelocator.GetPlatform(),
	servicelocator.GetBricksIndex(),
	servicelocator.GetServicesIndex(),
	servicelocator.GetDockerClient(),
	servicelocator.GetModelsIndex(),
	orchestrator.SystemInitOptions{
		OnlyDockerImages:    onlyImages,
		OnlyPlatformAndLibs: onlyPlatformAndLibraries,
	},
	newInitEventCallback(printEvent)
)
```

cmd/arduino-app-cli/system/system_helper.go
```diff
@@ -0,0 +1,101 @@
...
+				return
+			}
+			if last, ok := lastPct[e.Label]; ok && last == e.Percent {
+				return
+			}
+			lastPct[e.Label] = e.Percent
+		}
+		next(e)
+	}
+}
```
lucarin91, line 101: I would keep things simpler. Is this really needed?

cmd/arduino-app-cli/system/system_helper.go
```diff
@@ -0,0 +1,101 @@
...
+
+// String implements feedback.Result.
+func (e *initEvent) String() string {
+	if e.Type == initProgressEventType {
+		return fmt.Sprintf("%s: %d%%", e.Label, e.Percent)
+	}
+	return e.Message
+}
+
+func newInitEvent(e orchestrator.InitEvent) *initEvent {
```
lucarin91: I will call this
```suggestion
func fromInitEvent(e orchestrator.InitEvent) *initEvent {
```

cmd/arduino-app-cli/system/system_helper.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+const (
+	initLogEventType      = "log"
+	initProgressEventType = "progress"
+)
+
+var _ feedback.Result = (*initEvent)(nil)
+
+type initEvent struct {
```
lucarin91: I would call this
```suggestion
type initResult struct {
```

cmd/feedback/feedback.go
```diff
@@ -22,16 +23,19 @@ type OutputFormat int
...
+	// JSONLines is a stream of compact JSON documents
+	JSONLines
 )
 
 var formats = map[string]OutputFormat{
-	"json":     JSON,
-	"jsonmini": MinifiedJSON,
-	"text":     Text,
+	"json":       JSON,
+	"jsonmini":   MinifiedJSON,
```
lucarin91: I would avoid exposing this if it is not needed

cmd/feedback/feedback.go
```diff
@@ -251,3 +256,34 @@ func PrintResult(res Result) {
...
 }
+
+// NewResultStream returns a function to print an unbounded stream of Results,
+// like the events emitted by a long running command. The results are written to
+// the out writer as they come, they are not accumulated, so this is the
+// counterpart of PrintResult for a stream (and the two must not be mixed).
+//
+// The JSON and MinifiedJSON formats are single-document formats, so they are
+// not available for a stream and the function errors: JSONLines must be used
+// instead, it renders the stream as one JSON object per line.
```
lucarin91: Try to make doc comments more direct; 3 lines are usually sufficient

internal/orchestrator/bricks/testdata/dummy-app-brick-override-temp/app.yaml
lucarin91, line 1: Not sure because this testdata file keeps changing; we should give that a look at some point

internal/orchestrator/docker.go
```diff
@@ -67,36 +66,61 @@ func parseDockerImage(image string) (name string, version string) {
 	return image, ""
 }
 
-// Returns the number of bytes that would be downloaded when pulling the new docker image while the old one is
-// already present locally. It accounts for image layers that are already present locally.
-func GetBytesToDownload(localRefStr string, remoteRefStr string) (int64, error) {
+// imageDisplayName returns a short, human-friendly name for a docker image
+// reference, e.g. "python-apps-base" for
+// "ghcr.io/arduino/app-bricks/python-apps-base:0.11.0rc6".
+func imageDisplayName(image string) string {
```
lucarin91: Not sure if the last bit is ok; The real name of the image is the whole `ghcr.io/arduino/app-bricks/python-apps-base`, I think it is correct to show that

internal/orchestrator/system.go
```diff
@@ -138,39 +145,80 @@ func downloadSupportedImages(ctx context.Context, cfg config.Configuration, bric
...
 			return err
 		}
 
 		// Check that there is enough disk space for the additional layers needed by the image.
-		previousExistingImage := GetHighestVersion(image, pulledImages)
-		if toDownload, err := GetBytesToDownload(previousExistingImage, image); err != nil {
-			// In case of errors getting the size to download, proceed anyway.
-			slog.Warn("Unable to get the new image layers size", "image", image, "error", err)
-		} else if uint64(float64(toDownload)*2.5) > freeSpace {
+		if uint64(float64(img.bytes)*2.5) > freeSpace {
```
lucarin91: I am not sure about this check here. I think here we aren't considering the shared layers among different images, which you instead evaluate in `totalBytes`. Shouldn't we do a single freeSpace check on that size?


```suggestion
		if uint64(float64(totalBytes)*2.5) > freeSpace {
```

internal/orchestrator/system.go
```diff
@@ -138,39 +145,80 @@ func downloadSupportedImages(ctx context.Context, cfg config.Configuration, bric
...
 	imagesToPreinstall = slices.DeleteFunc(imagesToPreinstall, func(v string) bool {
 		return slices.Contains(pulledImages, v)
 	})
 
+	type imageToPull struct {
+		ref   string
+		bytes int64
+	}
+	imagesToPull := make([]imageToPull, 0, len(imagesToPreinstall))
+	layersPerImage := make([][]dockerImageLayer, 0, len(imagesToPreinstall))
```
lucarin91: It doesn't seem that you need to collect layers per image, but just layers
```suggestion
	allLayers := make([]dockerImageLayer, 0, len(imagesToPreinstall))
```

internal/orchestrator/system.go
```diff
@@ -138,39 +145,80 @@ func downloadSupportedImages(ctx context.Context, cfg config.Configuration, bric
...
+		layers, err := missingLayers(previousExistingImage, image)
+		if err != nil {
+			slog.Warn("Unable to get the new image layers size", "image", image, "error", err)
+		}
+		var bytes int64
+		for _, l := range layers {
+			bytes += l.Size
+		}
+		imagesToPull = append(imagesToPull, imageToPull{ref: image, bytes: bytes})
+		layersPerImage = append(layersPerImage, layers)
```
lucarin91:
```suggestion
		allLayers = append(layersPerImage, layers...)
```

internal/orchestrator/system.go
```diff
@@ -138,39 +145,80 @@ func downloadSupportedImages(ctx context.Context, cfg config.Configuration, bric
...
+		slog.Info("docker image to download", "image", image, "bytes", bytes)
+	}
+	totalBytes := sumUniqueLayers(layersPerImage)
```
lucarin91:
```suggestion
	totalBytes := sumUniqueLayers(allLayers)
```

internal/orchestrator/system.go
```diff
@@ -203,20 +251,25 @@ func pullImage(ctx context.Context, stdout io.Writer, docker dockerClient.APICli
...
-				fmt.Fprintf(stdout, "%s", payload.Status)
-			}
-			if payload.Progress != "" {
-				fmt.Fprintf(stdout, "[%s] %s\r", payload.ID, payload.Progress)
-			} else {
-				fmt.Fprintln(stdout)
+			// Accumulate the downloaded bytes across all layers/images and report
+			// the global download progress.
+			downloaded := updateLayerProgress(layerProgress, payload.Status, payload.ID, payload.ProgressDetail.Current)
+			if totalBytes > 0 && eventCB != nil {
```
lucarin91: I would add to the top the following so we are sure there is always a callback
```suggestion
			if eventCB == nil { eventCB = func(event InitEvent) {}}
```

cmd/arduino-app-cli/system/system_helper.go
mirkoCrobu: It solves the following: 
Without it we have several times more events (7×+ in my test)
I guess it's worth keeping it. What do you think? 

image

cmd/feedback/feedback.go
mirkoCrobu: ok done

cmd/arduino-app-cli/system/system_helper.go
mirkoCrobu: done

cmd/arduino-app-cli/system/system_helper.go
mirkoCrobu: done

cmd/arduino-app-cli/system/system.go
mirkoCrobu: done

internal/orchestrator/system.go
mirkoCrobu: done

internal/orchestrator/system.go
mirkoCrobu: done

internal/orchestrator/system.go
mirkoCrobu: done

internal/orchestrator/system.go
mirkoCrobu: done

internal/orchestrator/system.go
mirkoCrobu: You're right!! Done

cmd/arduino-app-cli/system/system_helper.go
lucarin91: Who is generating all of those events? Is Docker itself?

mirkoCrobu: yes, it is Docker. It emits many events per second (for each layer) about the bytes variation, but we have an internal percentage that needs N bytes to increment by 1%. In the meantime, for each Docker event, we print the percentage. This function collapses the events and emits the events only if the percentage changes.

cmd/feedback/feedback.go
lucarin91: minified json is still here

internal/orchestrator/system.go
```diff
@@ -45,13 +45,36 @@ var ErrDockerOutOfSpace = errors.New("not enough disk space to pull the docker i
...
+	InitLogEvent InitEventType = iota
+	InitProgressEvent
+)
+
+type InitEventSource string
+
+const (
+	InitSourceDocker   InitEventSource = "docker"
+	InitSourceLibs     InitEventSource = "libs"
+	InitSourcePlatform InitEventSource = "platform"
```
lucarin91: `platform` here is misleading; usually that refers to the `arduino:zephyr` core (or platform). Instead, you used that for the installation of the `arduino-unoq` or `arduino-ventunoq` package. 

Maybe we should cluster them in 
- `docker` for the Docker images
- `arduino` for Arduino platform and libs
- `linux` or `deb` for the Debian packages

lucarin91 (approved):

cmd/feedback/feedback.go
mirkoCrobu: done

internal/orchestrator/system.go
mirkoCrobu: done

PR #490, opened 2026-07-03
feat: allow to add external unit socket

lucarin91:
### Motivation

We should have a general way to specify external service dependencies required by apps.

### Change description

With the new `ARDUINO_APP_CLI__REQUIRED_UNITS` we can specify a list of service, and the app-cli will auto-discover run folder either in `/run` or `/var/run`

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/config/config.go
+++ b/internal/orchestrator/config/config.go
@@ -25,7 +25,7 @@ var RunnerVersion = "0.11.0rc7"
 type Configuration struct {
 	appsDir                          *paths.Path
 	dataDir                          *paths.Path
-	requiredUnits                    []string
+	requiredRuntimes                 []string
 	customModelsDir                  *paths.Path
 	modelsDir                        *paths.Path
 	dockerRegistryBase               string
@@ -62,14 +62,14 @@ func NewFromEnv() (Configuration, error) {
 	}
 
 	// Required host units bind-mounted as /run/<unit> into app containers.
-	requiredUnitsEnv, ok := os.LookupEnv("ARDUINO_APP_CLI__REQUIRED_UNITS")
+	requiredRuntimesEnv, ok := os.LookupEnv("ARDUINO_APP_CLI__REQUIRED_RUNTIMES")
 	if !ok {
-		requiredUnitsEnv = "arduino-router,arduino-app-cloud"
+		requiredRuntimesEnv = "arduino-router,arduino-app-cloud"
 	}
-	var requiredUnits []string
-	for u := range strings.SplitSeq(requiredUnitsEnv, ",") {
+	var requiredRuntimes []string
+	for u := range strings.SplitSeq(requiredRuntimesEnv, ",") {
 		if u = strings.TrimSpace(u); u != "" {
-			requiredUnits = append(requiredUnits, u)
+			requiredRuntimes = append(requiredRuntimes, u)
 		}
 	}
 
@@ -133,7 +133,7 @@ func NewFromEnv() (Configuration, error) {
 	c := Configuration{
 		appsDir:                          appsDir,
 		dataDir:                          dataDir,
-		requiredUnits:                    requiredUnits,
+		requiredRuntimes:                 requiredRuntimes,
 		customModelsDir:                  customModelsDir,
 		modelsDir:                        modelsDir,
 		dockerRegistryBase:               registryBase,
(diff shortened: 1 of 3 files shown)


internal/orchestrator/config/config.go
```diff
@@ -126,7 +133,7 @@ func NewFromEnv() (Configuration, error) {
 	c := Configuration{
 		appsDir:                          appsDir,
 		dataDir:                          dataDir,
-		routerSocketPath:                 routerSocket,
+		requiredUnits:                    requiredUnits,
```
martacarbone: why unit instead of socket? runtime?

internal/orchestrator/config/config.go
```diff
@@ -172,8 +179,31 @@ func (c *Configuration) ExamplesDir() *paths.Path {
...
+// RequiredUnitsPaths returns the discovered host paths for configured required
+// units, searching in order: /run/<unit>, /var/run/<unit>, /run/<unit>.sock,
+// /var/run/<unit>.sock. The first existing entry per unit is returned.
+func (c *Configuration) RequiredUnitsPaths() paths.PathList {
+	var result paths.PathList
+	for _, unit := range c.requiredUnits {
+		candidates := []*paths.Path{
+			paths.New("/run", unit),
+			paths.New("/var/run", unit),
+			paths.New("/run", unit+".sock"),
```
mirkoCrobu: `/var/run/` is a symlink to `/run`, so what we pass to the container is `/run/arduino-router.sock` instead of `/var/run/arduino-router.sock.` Could this be a problem?

internal/orchestrator/provision.go
```diff
@@ -338,12 +338,12 @@ func generateMainComposeFile(
...
-	slog.Debug("Adding UNIX socket", slog.Any("sock", cfg.RouterSocketPath().String()), slog.Bool("exists", cfg.RouterSocketPath().Exist()))
-	if cfg.RouterSocketPath().Exist() {
+
+	for _, unitPath := range cfg.RequiredUnitsPaths() {
 		volumes = append(volumes, volume{
 			Type:   "bind",
-			Source: cfg.RouterSocketPath().String(),
-			Target: "/var/run/arduino-router.sock",
+			Source: unitPath.String(),
+			Target: unitPath.String(),
```
giulio93: Does the container expect the socket to be always in `/var/run/arduino-router.sock`, or is dinamically set inside it?

internal/orchestrator/config/config.go
lucarin91: Yeah, probably `requiredRuntimes` is better

internal/orchestrator/provision.go
lucarin91: The first one https://github.com/arduino/app-bricks-py/blob/34ef6daee8051a64c7f917e6a4a43ec8c5c48988/src/arduino/app_utils/bridge.py#L283

internal/orchestrator/config/config.go
lucarin91: Apparently, the same symlink is also inside the Python container, so that works. I think this is a pretty common thing on Linux distros, so maybe not a problem, and anyway we were thinking to move the router socket in /run at some point

internal/orchestrator/config/config.go
```diff
@@ -172,8 +179,31 @@ func (c *Configuration) ExamplesDir() *paths.Path {
...
+		for _, p := range candidates {
+			if p.Exist() {
+				result.AddIfMissing(p)
+				found = true
+				break
+			}
+		}
+		if !found {
+			slog.Debug("required unit not found on host", "unit", unit)
+		}
```
martacarbone:
```suggestion
	for _, runtime := range c.requiredRuntimes {
		candidates := []*paths.Path{
			paths.New("/run", runtime),
			paths.New("/var/run", runtime),
			paths.New("/run", runtime+".sock"),
			paths.New("/var/run", runtime+".sock"),
		}
		found := false
		for _, p := range candidates {
			if p.Exist() {
				result.AddIfMissing(p)
				found = true
				break
			}
		}
		if !found {
			slog.Debug("required runtime not found on host", "runtime", runtime)
		}
```

martacarbone (approved):

mirkoCrobu (approved):

fabik111:
Tested with the new Arduino cloud daemon, and it worked, thank you!

PR #494, opened 2026-07-06
fix: some service images were not downloaded

lucarin91:
### Motivation

The  system init  command was failing to download some brick service Docker images. There were two issues:

1.  imagePrefixes  was missing some newly added image prefixes.
2. Services required by bricks without a Compose file were never resolved, so their images were skipped.

### Change description

- update the `imagesPrefixex` with codelinaro images
- fix the above-mentioned bug
- added a test for future regression

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/system.go
+++ b/internal/orchestrator/system.go
@@ -242,7 +242,13 @@ func isTemporaryDockerError(err error) bool {
 }
 
 // List of prefixes used to identify current or past Arduino images. Used both during 'system init' and during cleanup.
-var imagePrefixes = []string{"ghcr.io/bcmi-labs/", "public.ecr.aws/arduino/", "ghcr.io/arduino/", "influxdb"}
+var imagePrefixes = []string{
+	"ghcr.io/bcmi-labs/",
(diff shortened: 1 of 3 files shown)


internal/orchestrator/system.go
```diff
@@ -370,17 +376,22 @@ func SystemCleanup(ctx context.Context, cfg config.Configuration, bricksindex *b
 	}
 
 	// Remove unused images
-	containersMustStay, err := getRequiredImages(cfg, bricksindex, servicesindex, modelsIndex)
+	imagesMustStay, err := getRequiredImages(cfg, bricksindex, servicesindex, modelsIndex)
 	if err != nil {
 		return result, err
 	}
+	slog.Debug("images that must stay", "imagesMustStay", imagesMustStay)
```
Xayton, line 383: Do you want to keep the slog.Debug messages? Or did you forget to remove after debugging?

Xayton (approved):

internal/orchestrator/system.go
lucarin91: I was thinking to keep it

PR #501, opened 2026-07-06
Update event inconsistency

### LINKED ISSUE #382: update events inconsistency output
We return Done event only if the apt package is correctly updated, otherwise we return an error. But if the platform fails to upload, we return an event error and also a done event

We should decide if the error event should be considered as the end of the upgrade operation or if we should use the done instead.

See also: https://github.com/arduino/arduino-app-cli/issues/381

mirkoCrobu:
### Motivation

closes #382
When the apt upgrade failed, the stream stopped after the `error` event without
emitting any terminal event, so clients had no reliable way to know the
operation had ended.

### Change description

- `error` events no longer end the operation
- `done` is the terminal event, always emitted last, also after errors
- If the upgrade restarts `arduino-app-cli` itself, the stream can be cut before
`done` is delivered. in this case, `restart` is the last event
- SSE documentation and tests updated

### Reviewer checklist

### DIFF
--- a/cmd/gendoc/docs.go
+++ b/cmd/gendoc/docs.go
@@ -1030,6 +1030,9 @@ Contains a JSON object with the details of an error.
 				ContentType:   "text/event-stream",
 				DataStructure: "",
 				Description: `A stream of Server-Sent Events (SSE) that notifies the progress of the update process.
+The 'done' event is the only terminal event: an 'error' reports a step that failed but does
+not end the operation, so the client should collect the errors received during the stream and
+present a final summary once 'done' arrives.
 The client will receive events formatted as follows:
 
 **Event 'log'**:
@@ -1054,9 +1057,16 @@ Contains a string with the message that the upgrade is completed and the system 
 'data: Upgrade completed. Restarting'
 
 **Event 'error'**:
-Contains a JSON object with the details of an error.
+Contains a JSON object with the details of a step that failed. It does not end the operation:
+the upgrade continues and 'done' is emitted anyway.
 'event: error'
 'data: {"code":"internal_service_err","message":"An error occurred during operation"}'
+
+**Event 'done'**:
+Contains a string with the message that the update process is complete. It is always emitted
+last, whether or not 'error' events were received.
+'event: done'
+'data: Update completed'
 `,
 			},
 			PossibleErrors: []ErrorResponse{
--- a/cmd/gendoc/docs.go
+++ b/cmd/gendoc/docs.go
@@ -1030,7 +1030,7 @@ Contains a JSON object with the details of an error.
 				ContentType:   "text/event-stream",
 				DataStructure: "",
 				Description: `A stream of Server-Sent Events (SSE) that notifies the progress of the update process.
-The 'done' event is the only terminal event: an 'error' reports a step that failed but does
+The 'done' and 'restarting' events are the only terminal event: an 'error' reports a step that failed but does
 not end the operation, so the client should collect the errors received during the stream and
 present a final summary once 'done' arrives.
 The client will receive events formatted as follows:
@@ -1063,8 +1063,8 @@ the upgrade continues and 'done' is emitted anyway.
 'data: {"code":"internal_service_err","message":"An error occurred during operation"}'
 
 **Event 'done'**:
-Contains a string with the message that the update process is complete. It is always emitted
-last, whether or not 'error' events were received.
+Contains a string with the message that the update process is complete. It is emitted last,
+also when 'error' events were received. 
 'event: done'
 'data: Update completed'
 `,
--- a/internal/update/update_test.go
+++ b/internal/update/update_test.go
@@ -189,7 +189,7 @@ func TestManagerUpgradePackages(t *testing.T) {
 		packages   []UpgradablePackage
 		svc1Err    error
 		svc2Err    error
-		expectEvts []EventType // expected last event type from broadcast
+		expectEvts []EventType // expected event sequence from broadcast
 	}{
 		{
 			name:       "both services upgrade successfully",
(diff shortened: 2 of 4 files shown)


mirkoCrobu:
### TEST

Insert a fake warning event: 		
`m.broadcast(NewWarningEvent(fmt.Sprintf("failed to upgrade Arduino packages:")))`

**on the board:** 
curl http://localhost:8800/v1/system/update/events
then  curl -X PUT http://localhost:8800/v1/system/update/apply

App Lab could collect warnings and show them to the user at the end or durign the streaming

image

internal/update/update_test.go
```diff
@@ -200,13 +200,13 @@ func TestManagerUpgradePackages(t *testing.T) {
...
 			svc1Err:    errors.New("arduino upgrade failed"),
-			expectEvts: []EventType{ErrorEvent, DoneEvent}, // should continue to deb upgrade and complete
+			expectEvts: []EventType{WarningEvent, DoneEvent},
 		},
 		{
 			name:     "deb service fails",
 			packages: []UpgradablePackage{{Type: Debian, Name: "pkg1", ToVersion: "1.1"}},
 			svc2Err:  errors.New("deb upgrade failed"),
-			// FIXME: we should alwas return Done?
+			// deb failure is blocking: the operation ends with an error and no done event.
```
lucarin91: I think we are always expecting a done event; check here, for instance https://github.com/arduino/arduino-app-cli/blob/f004fdc00b3f7a29422834a9dc3752537e5dbd77/cmd/arduino-app-cli/system/system.go#L126

lucarin91 (changes requested):
I like the idea of adding a non blocking warning event, but I would verify also the semantic of the error event; it seems that in some places we consider an error event a synonym of done with error, in other places we aren't.

internal/update/update_test.go
mirkoCrobu: right. Since the error event is terminal (not followed by a done event), we could make it stop the event loop with a break too

cmd/arduino-app-cli/system/system.go
```diff
@@ -123,7 +123,9 @@ func newUpdateCmd(cfg config.Configuration) *cobra.Command {
 					feedback.Printf("[%s] %s", event.Type.String(), event.GetData())
 				}
 
-				if event.Type == update.DoneEvent {
+				// ErrorEvent is terminal (done with error): it is not followed
+				// by a DoneEvent, so we must stop consuming events here too.
+				if event.Type == update.DoneEvent || event.Type == update.ErrorEvent {
```
Xayton: I think this is not true because `UpgradePackages`, when returning a RestartEvent in the defer section, could also generate an error event. 

But this might be related to the idea of moving up the selfUpgrade.
Alternative fixes:

1. Use a (newly added) warning event instead of error event, if the restart fails
2. Use a named return, and change the returned value if the restart fails, in addition to broadcasting an error event. This way the caller will know something went wrong from the return value of `UpgradePackages`.

internal/update/event.go
```diff
@@ -5,7 +5,9 @@
 
 package update
 
-import "go.bug.st/f"
+import (
+	"go.bug.st/f"
+)
```
Xayton: Useless change?

cmd/arduino-app-cli/system/system.go
mirkoCrobu: Thanks, you're right. Now all errors and warnings are non-blocking, and done is always the last event emitted.

About moving up the selfUpgrade, we're tracking it here: https://github.com/arduino/arduino-app-cli/issues/521

lucarin91 (changes requested):
you dind't wire the internal update function to the endpoint and the cli

mirkoCrobu:
> you dind't wire the internal update function to the endpoint and the cli

warning falls into the default case. 
But making it explicit can improve readability.

lucarin91:
>all errors are non-blocking

now that have warning why errors are non-blocking

mirkoCrobu:
> > all errors are non-blocking
> 
> now that have warning why errors are non-blocking

Today, App Lab never treats errors as terminal. Changing the error behaviours now would break that, and the new App Lab wouldn't be able to update old boards.

So we agreed on: the CLI always emits done as the only terminal event, and App Lab collects warnings and errors during the stream to show a single "done with errors/warnings" summary at the end.

lucarin91:
> > > all errors are non-blocking
> > 
> > 
> > now that have warning why errors are non-blocking
> 
> Today, App Lab never treats errors as terminal. Changing the error behaviours now would break that, and the new App Lab wouldn't be able to update old boards.
> 
> So we agreed on: the CLI always emits done as the only terminal event, and App Lab collects warnings and errors during the stream to show a single "done with errors/warnings" summary at the end.

If that is the case, I would avoid adding a warning event; what is the advantage then?

mirkoCrobu:
> > > > all errors are non-blocking
> > > 
> > > 
> > > now that have warning why errors are non-blocking
> > 
> > 
> > Today, App Lab never treats errors as terminal. Changing the error behaviours now would break that, and the new App Lab wouldn't be able to update old boards.
> > So we agreed on: the CLI always emits done as the only terminal event, and App Lab collects warnings and errors during the stream to show a single "done with errors/warnings" summary at the end.
> 
> If that is the case, I would avoid adding a warning event; what is the advantage then?

At this point I agree with you. Nothing in the update flow is really a warning: we just have real failures.
We can add the event type later if a warning case shows up.

So, done is always emitted as the only terminal event.

lucarin91:
>done event is always sent 

This is also not true. The apt update internally triggers an app-cli daemon restart, which will kill the process itself without sending any done event. So at most we return done also in case of error.

cmd/gendoc/docs.go
```diff
@@ -1030,6 +1030,9 @@ Contains a JSON object with the details of an error.
 				ContentType:   "text/event-stream",
 				DataStructure: "",
 				Description: `A stream of Server-Sent Events (SSE) that notifies the progress of the update process.
+The 'done' event is the only terminal event: an 'error' reports a step that failed but does
```
lucarin91:
```suggestion
The 'done' and 'restarting' events are the only terminal event: an 'error' reports a step that failed but does
```

mirkoCrobu:
> > done event is always sent
> 
> This is also not true. The apt update internally triggers an app-cli daemon restart, which will kill the process itself without sending any done event. So at most we return done also in case of error.

I would handle the restart issue in another task, moving it to the manager level.  I think it is the natural place to trigger the restart, and it will fix the done issue. 

But for now, we can go with your proposal, and I will document that done is emitted also in case of errors.

lucarin91:
> > > done event is always sent
> > 
> > 
> > This is also not true. The apt update internally triggers an app-cli daemon restart, which will kill the process itself without sending any done event. So at most we return done also in case of error.
> 
> I would handle the restart issue in another task, moving it to the manager level. I think it is the natural place to trigger the restart, and it will fix the done issue.
> 
> But for now, we can go with your proposal, and I will document that done is emitted also in case of errors.

That's ok but you need to update the PR description then

lucarin91:
@mirkoCrobu can you also update the PR description?

lucarin91 (approved):

PR #505, opened 2026-07-06
fix: stop failed apps before starting a new one

### LINKED ISSUE #495: service cleanup seems to not work correctly cross application
If an application crashes it could leave behind some service running, those service could make other application that requires those service fails.

We currently have in place a clean up feature that remove service of an app before running it, to avoid conflicts, but this seems to not work across application.

lucarin91:
### Motivation

An app in `Failed` status can still have some of its containers running (`Failed` means at least one container failed, not that all are stopped). Those leftover containers can interfere with launching apps that use the same services.

### Change description

`StartApp` now stops any *other* app in `Failed` before starting (best-effort; errors logged as warnings). The app being started is skipped, so retrying a failed app still works.

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -104,6 +104,10 @@ func StartApp(
 	verbose bool,
 	cb func(StreamMessage),
 ) error {
+	if cb == nil {
+		cb = func(StreamMessage) {}
+	}
+
 	bricksIndex = bricksIndex.WithAppBricks(appToStart.LocalBricks)
 
 	if err := checkBricks(appToStart.Descriptor, bricksIndex, modelsIndex); err != nil {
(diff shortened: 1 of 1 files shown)


Xayton:
Before merging this, we need to wait for [this PR](https://github.com/arduino/app-bricks-py/pull/346) to be merged and released.

UPDATE: the PR above has been merged and released in https://github.com/arduino/app-bricks-py/releases/tag/release%2F0.11.0rc9

internal/orchestrator/orchestrator.go
```diff
@@ -119,14 +123,39 @@ func StartApp(
...
+		case StatusFailed:
+			if s.AppPath.EqualsTo(appToStart.FullPath) {
+				continue
+			}
+			failedApp, err := app.Load(s.AppPath)
+			if err != nil {
+				slog.Warn("skipping failed app: cannot load", slog.String("path", s.AppPath.String()), slog.String("error", err.Error()))
+				continue
+			}
+			if err := StopApp(ctx, docker, platform, failedApp, nil); err != nil {
```
Xayton, line 152: I would add a debug log here, logging that we are stopping a failed app, for the reasons explained in the comment above. This could help debug issues when starting apps.

Right now we log only if stopping the app failed.

Xayton (approved):

Xayton (approved):

internal/orchestrator/orchestrator.go
```diff
@@ -119,14 +123,40 @@ func StartApp(
...
+		case StatusRunning, StatusStarting:
+			runningApp, err := app.Load(s.AppPath)
+			if err != nil {
+				return fmt.Errorf("failed to load running app: %w", err)
+			}
+			return fmt.Errorf("app %q is running", runningApp.Name)
+		case StatusFailed:
+			if s.AppPath.EqualsTo(appToStart.FullPath) {
+				continue
+			}
```
dido18, line 145: we could always stop the app even if the app to start is the same

Xayton: I thought, in that case, starting the same app will take care of closing anything left by the previous run. But if that is not the case, then yes.

PR #508, opened 2026-07-07
fix: ai icon from brick list is not shown

dido18:
### Motivation
The following Bricks don't display the "AI" icon:
- Gesture Recognition
- Large Language Model (LLM)
- Object Detection
- Video Object Detection

Expected
 All Bricks with AI models should have the AI icon.

### Change description
The brick is an AiBrick is resolved at runtime by app-cli by looking at the `model_name` and the `model_by_board` filed in the brick list. The `require_model` filed in the yaml is removed.

ai brick: if the "model_name" is not empty or the "model_by_board" is not empty.
Precedence to the `model_name`

### Additional Notes
-- https://github.com/arduino/app-bricks-py/pull/348 

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/bricksindex/bricks_index.go
+++ b/internal/orchestrator/bricksindex/bricks_index.go
@@ -220,7 +220,7 @@ func (b Brick) GetModelNameByBoard(platform platform.Platform) string {
 	return defaultModelName
 }
 
-func (b Brick) isModelRequired(platform platform.Platform) bool {
+func (b Brick) isAiBrick(platform platform.Platform) bool {
 	for _, mb := range b.ModelByBoard {
 		if mb.Platform == platform.BoardName && mb.Model != "" {
 			return true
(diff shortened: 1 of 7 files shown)


dido18:
tested on UNOQ

image

tested on Ventunoq

image

internal/orchestrator/bricksindex/bricks_index.go
```diff
@@ -220,6 +220,18 @@ func (b Brick) GetModelNameByBoard(platform platform.Platform) string {
...
+	if b.ModelName != "" {
+		return true
+	}
+	for _, mb := range b.ModelByBoard {
+		if mb.Platform == platform.BoardName && mb.Model != "" {
+			return true
+		}
+	}
+	return false
+}
```
lucarin91: Not sure why we are doing this at runtime. I think we could set `ModelName` to the correct value during model index loading; the platform doesn't change.

internal/orchestrator/bricksindex/bricks_index.go
```diff
@@ -220,6 +220,18 @@ func (b Brick) GetModelNameByBoard(platform platform.Platform) string {
 	return defaultModelName
 }
 
+func (b Brick) isModelRequired(platform platform.Platform) bool {
```
martacarbone: we could write
func (b Brick) isModelRequired(platform platform.Platform) bool {
	return b.GetModelNameByBoard(platform) != ""
}

dido18: I prefer to keep the logic separated (even if they currenlty match). 
I renamed the funciotn into `isAiBrick`  to convey the fact that this logic is to check if the brick is an AI brick and could be extended with other field (see ai_framework_compatibility)

lucarin91 (approved):

PR #514, opened 2026-07-07
fix: compose file parsing in system init

Xayton:
### Motivation

Fix for the error while doing the upgrade (which calls system init).
```
failed to download container images used in examples: validating : error in parsing "compose-spec.json": getwd: no such file or directory
```

### Change description

This code was not executed before the PR linked below, so the bug was there but not visible.
The fix is to avoid doing a validation of the compose file while calling `loader.LoadWithContext`. This happens when there is no current working dir set (for example during .deb postinst).

See also the corresponding `loader.LoadWithContext` inside `logs.go`.

### Additional Notes

This bug was triggered by this change:
- https://github.com/arduino/arduino-app-cli/pull/494


martacarbone (approved):
lgtm

PR #519, opened 2026-07-09
fix(update): make sure that all service will be restarted

### LINKED ISSUE #521: Improve and fix the service restart logic
## Current status
When the arduino-app-cli triggers an update, it calls apt to perform the download, update of the .deb packages and then the restart of the services that need it. The only problem is that it must not restart arduino-app-cli daemon, otherwise the upgrade will be interrupted.
So the upgrade code asks apt to not restart services and that will be done at the end, manually, by arduino-app-cli, using the needrestart tool.

## Current issues
- This approach is not reliable, and needsrestat has a different behaviour on Ubuntu (due to [this patch](https://git.launchpad.net/ubuntu/+source/needrestart/plain/debian/patches/lp2067482/0002-defer-restarting-the-service-that-called-needrestart.patch)).
- In some cases the services that need to be restarted are not restarted in order (so `arduino-app-cli` might be restarted before for example `arduino-router`, and then it terminates, including the needrestart process, skipping the restart of the remaining services).
- After this fix https://github.com/arduino/arduino-app-cli/pull/518 there is an imperfection in which the update over cli command will return a not 0 exit code

Link to existing PRs:
- https://github.com/arduino/arduino-app-cli/pull/518
- https://github.com/arduino/arduino-app-cli/pull/519

## How to test the upgrade:
```
curl -X PUT http://localhost:8800/v1/system/update/apply?only-arduino=true && curl -s http://localhost:8800/v1/system/update/events
```

lucarin91:
### Motivation

The service restart operation could kill the `arduino-app-cli` daemon before the restart of the other service is really executed, which will create a situation in which, after an update, not every service runs the latest installed version.

To avoid that, a possible solution is to use an approach similar to the need-restart version patched by Ubuntu, in which the command excludes the service currently running the update.

### Change description

1. Ensure need-restart runs during apt upgrade by adding the env variable in the sudoers
2. Remove the explicit need-restart call after the update
3. Add a need-restart config to exclude the current running service from the restart operation
4. Avoid cli to self kill during update

### Reviewer checklist


lucarin91:
## Tests to do on Debian

The following tests could be run via the check update CI, or by a slight modification of that. I am not sure the test `New Version to new new version (via daemon)` is executed there. The important part is to verify that the update did exactly the below steps

### Current version to new version (via daemon)

What should happen:

1. During the apt upgrade, everything updates, but nothing is restarted
2. The new version executes needrestart -r a, which should restart everything else except the current service
3. The current daemon exits with SIGINT
 
### New Version to new new version (via daemon)

What should happen:

1. During the apt upgrade, everything updates and restarts except for the current service
2. The new new version SIGINT

mirkoCrobu:
### Manual test on UNO Q


1. board running the PR build. 
2. The same build was repackaged as `99.9.9` and served from a local apt repo as the new target version. 
3. `curl -X PUT http://localhost:8800/v1/system/update/apply`
4. Result: arduino-app-cli `0.0.0-9babeb739` to `99.9.9`:

**From the logs:** 
- apt ran from `15:04:23` to `15:05:14`; `journalctl -u arduino-app-cli` restricted to that window returns `-- No entries --`, i.e. needrestart did not restart the calling unit
- the upgrade completed and the daemon stayed alive through it.
- it shut itself down only at `15:15:03`. Systemd brought it back at `15:15:06`, with `NRestarts=1`


- `arduino-app-cli version` now reports `99.9.9` for both the installed binary and the running daemon
image

mirkoCrobu (commented):
LGTM, tested on UnoQ

lucarin91:
> ### Manual test on UNO Q
> 1. board running the PR build.
> 2. The same build was repackaged as `99.9.9` and served from a local apt repo as the new target version.
> 3. `curl -X PUT http://localhost:8800/v1/system/update/apply`
> 4. Result: arduino-app-cli `0.0.0-9babeb739` to `99.9.9`:
> 
> **From the logs:**
> 
> * apt ran from `15:04:23` to `15:05:14`; `journalctl -u arduino-app-cli` restricted to that window returns `-- No entries --`, i.e. needrestart did not restart the calling unit
> * the upgrade completed and the daemon stayed alive through it.
> * it shut itself down only at `15:15:03`. Systemd brought it back at `15:15:06`, with `NRestarts=1`
> * `arduino-app-cli version` now reports `99.9.9` for both the installed binary and the running daemon
> 
> image

@mirkoCrobu this test only the case `New Version to new new version (via daemon)` We should test also the other case write here https://github.com/arduino/arduino-app-cli/pull/519#issuecomment-5368443926

mirkoCrobu:
### Manual test on UNO Q  current version to new version

1. Board to  `0.13.0`.
2. A `99.9.9` build of this branch served from a local apt repo as the new target version.
3. `curl -X PUT http://localhost:8800/v1/system/update/apply`
4. Result: arduino-app-cli `0.13.0` to `99.9.9`:

**From the logs:**
- apt ran from `15:17:47` to `15:18:14`; `journalctl -u arduino-app-cli`. Nothing was restarted during the upgrade
- the old daemon then ran `needrestart -r a` itself, and was skipped by the config file the new package had just installed, so it stayed alive
- it shut itself down only at `15:19:55`. Systemd brought it back at `15:19:58`, with `NRestarts=1`

image

mirkoCrobu (approved):

PR #524, opened 2026-07-10
chore: enable linter to spot unused parameter

dido18:
### Motivation
The linter was not able to detect unused parameter.

### Change description
Enabled with 
```
   unparam:
      check-exported: true
```
      
  Removed all not used params

### Reviewer checklist

### DIFF
--- a/cmd/arduino-app-cli/daemon/daemon.go
+++ b/cmd/arduino-app-cli/daemon/daemon.go
@@ -53,7 +53,6 @@ func NewDaemonCmd(cfg config.Configuration, version string) *cobra.Command {
 					servicelocator.GetServicesIndex(),
 					servicelocator.GetAppIDProvider(),
 					cfg,
-					servicelocator.GetStaticStore(),
 					servicelocator.GetPlatform(),
 				)
 				if err != nil {
@@ -107,7 +106,6 @@ func httpHandler(ctx context.Context, cfg config.Configuration, daemonPort, vers
 			arduino.NewArduinoPlatformUpdater(servicelocator.GetPlatform(), cfg.ArduinoPlatformVersionConstraint),
 		),
 		servicelocator.GetProvisioner(),
-		servicelocator.GetStaticStore(),
 		servicelocator.GetModelsIndex(),
 		servicelocator.GetBricksIndex(),
 		servicelocator.GetServicesIndex(),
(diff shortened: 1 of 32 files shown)


internal/orchestrator/provision.go
```diff
@@ -108,7 +108,7 @@ func NewProvision(
 }
 
 func (p *Provision) App(
-	ctx context.Context,
+	_ context.Context,
```
lucarin91: I would remove the ctx here

internal/orchestrator/orchestrator.go
```diff
@@ -726,7 +722,7 @@ type CreateAppResponse struct {
 }
 
 func CreateApp(
-	ctx context.Context,
+	_ context.Context,
```
lucarin91: I would remove the context also here

internal/orchestrator/orchestrator.go
```diff
@@ -772,7 +768,7 @@ type CloneAppResponse struct {
 }
 
 func CloneApp(
-	ctx context.Context,
+	_ context.Context,
```
lucarin91: also here

internal/orchestrator/archive.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 func ExportAppZip(
-	ctx context.Context,
+	_ context.Context,
```
lucarin91: and here

Xayton:
Why in some cases you kept the unused parameter, but just renaming it to `_`?
Was it not possible to remove it and update the call points?

cmd/arduino-app-cli/daemon/daemon.go
```diff
@@ -53,7 +53,6 @@ func NewDaemonCmd(cfg config.Configuration, version string) *cobra.Command {
 					servicelocator.GetServicesIndex(),
 					servicelocator.GetAppIDProvider(),
 					cfg,
-					servicelocator.GetStaticStore(),
```
lucarin91, line 56: Is this `StaticStore` used anywhere anymore?

lucarin91 (changes requested):

cmd/arduino-app-cli/daemon/daemon.go
dido18: I am looking if it can be removed as well

dido18: I think we can simplfy the StaticStore usage, currently it is only used to pass the Assets folder and the services folder. I think they can be moved inside the config. 

I can open a subsequent PR to clean it

lucarin91 (approved):

PR #525, opened 2026-07-10
chore: fix upgrade test after needrestart defer fix

### LINKED ISSUE #521: Improve and fix the service restart logic

lucarin91:
### Reviewer checklist

### DIFF
--- a/internal/e2e/updatetest/helpers.go
+++ b/internal/e2e/updatetest/helpers.go
@@ -9,6 +9,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"iter"
 	"net"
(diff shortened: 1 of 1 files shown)


mirkoCrobu (approved):

martacarbone (approved):
lgtm

dido18 (approved):

PR #526, opened 2026-07-10
Move down chown in postinst

### LINKED ISSUE #522: running app-cli as root creates folder/files as root
First of all we should verify if this is a bug or not. app-cli explicetelly prevent the user to run any command with root user, but we allow to override the check with `ALLOW_ROOT` env variable for special cases.

So if the user execute the cli in that way should we prevent to create files or should we leave the user to fixing any possible issue.

For instance some folder that are created are here in the config initialization https://github.com/arduino/arduino-app-cli/blob/e510c28f82d153a290220427e341daece1167299/internal/orchestrator/config/config.go#L166

If we decide this is not a bug we need to create a chmod arduino:arduino of /var/lib/arduino-app-cli here https://github.com/arduino/arduino-deb-images/blob/e7713f1797945296e4ab77dc7040b5d2c32aa047/debos-recipes/qualcomm-linux-debian-rootfs-arduino-extra.yaml#L262-L265

Xayton:
The postinst script will ensure all the folders in `/var/lib/arduino-app-cli` are owned by arduino.
This must be done after calling the arduino-app-cli for "completion bash".

### DIFF
--- a/debian/arduino-app-cli/DEBIAN/postinst
+++ b/debian/arduino-app-cli/DEBIAN/postinst
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-chmod -R 775 /var/lib/arduino-app-cli
-
 systemctl enable arduino-app-cli
 systemctl enable arduino-avahi-serial.service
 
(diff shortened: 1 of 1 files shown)


debian/arduino-app-cli/DEBIAN/postinst
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-chown -R arduino:arduino /var/lib/arduino-app-cli \
-|| chown -R :arduino /var/lib/arduino-app-cli
-
 chmod -R 775 /var/lib/arduino-app-cli
```
dido18: we can move below also this

dido18 (approved):

PR #528, opened 2026-07-10
fix: avoid creating folder in /var/lib for permission issue

### LINKED ISSUE #522: running app-cli as root creates folder/files as root

lucarin91:
### Motivation

Ensure we only create user folders during config parsing.

### Reviewer checklist


internal/orchestrator/archive.go
```diff
@@ -185,6 +185,9 @@ func ImportAppFromZip(
 		finalDestPath, _ = findAppPathByName(newName, cfg)
 	}
 
+	if err := finalDestPath.Parent().MkdirAll(); err != nil {
+		return app.ID{}, fmt.Errorf("unable to create apps directory: %w", err)
+	}
```
mirkoCrobu: This case is covered by `ensureUserDirs()` in both cases: cli and api. I guess it is not needed, or is it covering a usecase that I am not taking in account? maybe if the folder is removed at run time?

lucarin91: Yeah, it is a leftover, thanks

mirkoCrobu:
@lucarin91  I am wondering if we need a pre-check of all required folders by config.go to be aware about missing folders instead of finding it out just during some commands. What do you think?

internal/orchestrator/config/config.go
```diff
@@ -151,25 +146,30 @@ func NewFromEnv() (Configuration, error) {
...
 	}
-	if err := c.AssetDir().MkdirAll(); err != nil {
+
+	if err := c.AppsDir().MkdirAll(); err != nil {
 		return err
 	}
-	if err := c.ModelsDir().MkdirAll(); err != nil {
+	if err := c.CustomModelsDir().MkdirAll(); err != nil {
 		return err
 	}
```
dido18, line 170: we are changin the logic.
before the the mkdir fail a log.warn was produced without exiting.

internal/orchestrator/config/config.go
```diff
@@ -151,25 +146,30 @@ func NewFromEnv() (Configuration, error) {
...
-	if err := c.AppsDir().MkdirAll(); err != nil {
-		return err
+// EnsureFolders creates the folders in the $HOME directory and
+// verifies the shared folders under DataDir are present.
+func (c *Configuration) EnsureFolders() error {
+	required := []*paths.Path{
+		c.AssetDir().Parent(),
+		c.ModelsDir(),
 	}
-	if err := c.examplesDir().Join("common").MkdirAll(); err != nil {
```
dido18: why the `examplesDir().Join("common")` createion is removed ? not needed anymore ?

lucarin91: Yeah, I don't think we needed if there are no examples; we simply don't show any examples

internal/orchestrator/config/config.go
lucarin91: Yeah, the idea is to create all the folders that are in the user's home folder. If that fails, we just exit, isn't safe to continue

internal/orchestrator/config/config.go
```diff
@@ -151,33 +146,30 @@ func NewFromEnv() (Configuration, error) {
...
+		if d.NotExist() {
+			return fmt.Errorf("required directory %s does not exist: reinstall arduino-app-cli or create it manually (chown arduino:arduino)", d)
 		}
 	}
 
-	if err := c.AssetDir().MkdirAll(); err != nil {
+	if err := c.AppsDir().MkdirAll(); err != nil {
 		return err
 	}
-	if err := c.ModelsDir().MkdirAll(); err != nil {
```
mirkoCrobu, line 178: Who is creating modelsDir now?

martacarbone (approved):

internal/e2e/e2e.go
```diff
@@ -73,6 +73,7 @@ func NewArduinoAppCLI(t *testing.T, opts ...ArduinoAppCLIOption) *ArduinoAppCLI
 		require.NoError(t, os.CopyFS(dataDir.String(), os.DirFS(originalTestDataDir.String())))
 		require.NoError(t, err, "failed to copy testdata to temp dir")
 	}
+	require.NoError(t, dataDir.Join("models").MkdirAll())
```
mirkoCrobu, line 76: I guess you need to add require.NoError(t, appDir.MkdirAll()) like you did for archive.test and orchestrator_test to fix the test failing.

mirkoCrobu (approved):

PR #531, opened 2026-07-16
Support Media Carrier audio device via Pipewire

giulio93:
### Motivation

Apps that need audio via Media Carrier had no framework ensuring PipeWire is actually running when they start, or releasing it when they don't need it anymore.  UNO Q  in PC mode (headless) accessed via adb does not brings up `user@<uid>.service` (and therefore PipeWire/WirePlumber) on its own without a real login. 
On top of that, whatever mechanism starts it needs to coexist correctly with real logind sessions (SSH, lightdm) that might independently be using — or also relying on — the same PipeWire instance, without either side stepping on the other.

### Change description

- **Device detection** (`peripherals/devices.go`, `check.go`): detect when a `media-carrier` is configured (`HasCarrierSoundDevice`), and treat that as satisfying an app's microphone/speaker requirement alongside a physical ALSA device.
- **`internal/orchestrator/pipewire` package**:
  - `StartPipewire` waits for the user's D-Bus session bus to come up and enables/starts the PipeWire units (`pipewire.socket`, `pipewire-pulse.socket`, `wireplumber.service`).
  - `EnableLinger`/`DisableLinger` use logind's linger mechanism (`SetUserLinger` — needs no elevated privilege when applied to one's own uid) to keep `user@<uid>.service` running for as long as an app needs audio, regardless of an unrelated login session (SSH/lightdm) logging in and out in the meantime.
  - Ownership of the linger flag is tracked via a disk-persistent marker recording systemd's own linger-state file mtime at the moment it was enabled. This means:
    - if the daemon crashes or the board loses power while linger is on, it gets reconciled (turned back off) at the next daemon startup;
    - if a human independently enables linger themselves (`loginctl enable-linger`) while an app is running, the app stopping afterward never clobbers it.
- **`orchestrator.go`**: `StartApp`/`stopAppWithCmd` call `EnableLinger` + `StartPipewire` before starting an app that needs audio, and `DisableLinger` after stopping it.
- **`provision.go`**: bind-mounts the PipeWire socket (`/run/user/1000/pipewire-0`) into app containers.

1. Build and install on the board: `task board:install`.
2. Start an app that requires a media-carrier sound device, e.g. `arduino-app-cli app start user:<some-audio-app>` — confirm it starts successfully and audio actually plays.
3. While it's running, confirm linger got enabled: `loginctl show-user <uid> -p Linger` → `Linger=yes`.
4. SSH into the board and log back out again while the app keeps running. Confirm `user@<uid>.service` stays active (`systemctl status user@<uid>.service`) and audio is uninterrupted — this is the scenario linger exists to protect against.
5. Stop the app — confirm linger gets disabled again (`Linger=no`), unless another real session (e.g. that SSH session, if still open) is still active, in which case it should correctly stay on.
6. Edge case — independent human linger: start an app, then manually run `loginctl enable-linger <user>` while it's running, then stop the app. Confirm linger is **still on** afterward (not turned off by the app stopping).
7. Crash/reboot reconciliation: with an app running (linger on, owned by the daemon), kill the daemon process (or reboot the board) without stopping the app cleanly. On the next daemon startup, confirm the leftover linger gets reconciled (turned off), assuming nothing else needs it.


internal/orchestrator/pipewire/linger.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+const systemdLingerDir = "/var/lib/systemd/linger"
+
+func ownerPath(uid int) string {
+	return filepath.Join(os.Getenv("ARDUINO_APP_CLI__DATA_DIR"), "pipewire", fmt.Sprintf("linger-owner-%d", uid))
+}
+
+func setOwner(uid int, systemdMtime time.Time) error {
+	path := ownerPath(uid)
```
lucarin91, line 27: you should us `cfg.DataDir().Join("pipewire", fmt.Sprintf("linger-owner-%d", uid))`

lucarin91 (changes requested):
I am not sure that dbus is adding anything here. We are basically adding another depedency of app-cli, without gaining much. If we can do that without a root commands, I would prefer to do that instead.

Taskfile.yml
```diff
@@ -313,3 +313,52 @@ tasks:
...
+      BOARD_PATH: /home/arduino/arduino-app-cli-debug
+      DLV_PORT: '{{.DLV_PORT | default "2345"}}'
+    env:
+      GOTOOLCHAIN: local
+    cmds:
+      - CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -gcflags="all=-N -l" -o {{.DEBUG_BINARY}} ./cmd/arduino-app-cli
+      - adb push {{.DEBUG_BINARY}} {{.BOARD_PATH}}
+      - adb shell chmod +x {{.BOARD_PATH}}
+      - adb forward tcp:{{.DLV_PORT}} tcp:{{.DLV_PORT}}
+      - adb shell "DOCKER_REGISTRY_BASE=${DOCKER_REGISTRY_BASE:-ghcr.io/robgee86/} DOCKER_PYTHON_BASE_IMAGE=${DOCKER_PYTHON_BASE_IMAGE:-app-bricks/python-apps-base:dev-media-carrier-audio-support} /home/arduino/go/bin/dlv --listen=:{{.DLV_PORT}} --headless --api-version=2 exec {{.BOARD_PATH}} -- ${CLI_ARGS:-app list}"
```
lucarin91, line 364: I would add this in a separated PR

giulio93:
> I am not sure that dbus is adding anything here. We are basically adding another depedency of app-cli, without gaining much. If we can do that without a root commands, I would prefer to do that instead.

https://github.com/arduino/arduino-app-cli/pull/536

giulio93:
superseded by https://github.com/arduino/arduino-app-cli/pull/536

PR #545, opened 2026-07-26
Add support for new example layout

### LINKED ISSUE #535: New examples structure
### Describe the request

Update how example applications are organized and served.

### Describe the current behavior

The system exposes all examples in a single, unorganized flat list. In this setup, examples are served without structural categorization, or grouping.

### Arduino App CLI version

v0.12.1

### Additional context

_No response_

### Issue checklist

- [x] I searched for previous requests in [the issue tracker](https://github.com/arduino/arduino-app-cli/issues?q=)
- [x] I verified the feature was still missing when using the latest version
- [x] My request contains all necessary details

martacarbone:
### Motivation

Process others examples subdirectories inside /var/lib/arduino-app-cli/examples/, (core-and-foudation and bricks) excluding /inspirational/, which is already handled from legacy code.
Update the brickDetails API to search for brick examples in the brick examples folder.

Update tests.

### DIFF
--- a/internal/orchestrator/appid/id.go
+++ b/internal/orchestrator/appid/id.go
@@ -145,6 +145,16 @@ func (p *Provider) parseID(id string) (ID, error) {
 					break
 				}
 			}
+
+			pathPrefix, _, found := strings.Cut(appPath, "/")
+			if (found && pathPrefix == "core-and-foundational") ||
+				(found && pathPrefix == "bricks") {
+				path = p.cfg.ExamplesBaseDir().Join(appPath)
+				if path.Exist() {
+					break
+				}
+			}
+
 		default:
 			return ID{}, ErrInvalidID
 		}
--- a/internal/orchestrator/appid/id.go
+++ b/internal/orchestrator/appid/id.go
@@ -101,10 +101,22 @@ func (p *Provider) IDFromPath(path *paths.Path) (ID, error) {
 				if err != nil {
 					return ID{}, ErrInvalidID
 				}
-				id = "examples:" + rel.String()
+				id = "examples:" + "inspirational/" + rel.String()
 				isFromKnownLocation = true
 				isExample = true
 				break
+			}
+		}
+
+		for _, example := range p.cfg.ExamplesAdditionalDirs() {
+			if strings.HasPrefix(path.String(), example.String()) {
+				rel, err := path.RelFrom(p.cfg.ExamplesBaseDir())
+				if err != nil {
+					return ID{}, ErrInvalidID
+				}
+				id = "examples:" + rel.String()
+				isFromKnownLocation = true
+				isExample = true
 			}
 		}
 	}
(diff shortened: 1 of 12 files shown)


Xayton (changes requested):
Should run `go run ./cmd/gendoc`, the corresponding changes are missing in this PR.

internal/orchestrator/appid/id.go
Xayton, line 76: It looks like this function is not keeping into account the `ExampleAdditionalDirs`.

Is it correct that a brick example has a full path as the `encoded_id` (starting with `/var/lib/arduino-app-cli`)?

```sh
~$ curl -s "http://127.0.0.1:8800/v1/bricks/arduino:object_detection" | jq -r '.code_examples[] | {path, encoded_id, name}'
{
  "path": "/var/lib/arduino-app-cli/examples/bricks/arduino/object_detection/01_object_detection_example/python/main.py",
  "encoded_id": "L3Zhci9saWIvYXJkdWluby1hcHAtY2xpL2V4YW1wbGVzL2JyaWNrcy9hcmR1aW5vL29iamVjdF9kZXRlY3Rpb24vMDFfb2JqZWN0X2RldGVjdGlvbl9leGFtcGxl",
  "name": "Object Detection"
}
~$ curl -s http://127.0.0.1:8800/v1/examples | jq -r --arg b "arduino:object_detection" '.bricks[] | select(.brick==$b) | .examples[] | {id, encoded_id}'
{
  "id": "examples:bricks/arduino/object_detection/01_object_detection_example",
  "encoded_id": "ZXhhbXBsZXM6YnJpY2tzL2FyZHVpbm8vb2JqZWN0X2RldGVjdGlvbi8wMV9vYmplY3RfZGV0ZWN0aW9uX2V4YW1wbGU"
}
```

Unless I'm missing something, this function should also consider `ExamplesAdditionalDirs`.

Xayton (changes requested):
We should probably also update the `getUsedByApps` to exclude the brick examples.
In other words: when I open a brick, the brick examples should now show up in the list of `UsedByApps`.

internal/orchestrator/appid/id.go
martacarbone: Correct, the current version refers:
examples:inspirationals/ for legacy
examples:core-and-foundational and examples:bricks for other examples.

internal/orchestrator/orchestrator_test.go
```diff
@@ -554,7 +554,7 @@ func createApp(
 		require.NoError(t, err)
 		newID, err := idProvider.IDFromPath(newPath)
 		require.NoError(t, err)
-		assert.Empty(t, gCmp.Diff(f.Must(idProvider.ParseID("examples:"+name)), newID))
+		assert.Empty(t, gCmp.Diff(f.Must(idProvider.ParseID("examples:inspirational/"+name)), newID))
```
martacarbone, line 557: @Xayton This examples is using a legacy example, see line 522. This should have "examples:inspirational/" as id

internal/orchestrator/bricks/bricks.go
Xayton, line 302: I think here we should include core and foundational, but exclude brick examples.
Showing brick examples as apps using a given brick is redundant.

I would move the comment inside the function, close the the point where the pathsToExplore list.
```go
// Brick examples are not included here, because we do not want to show those as "apps using the brick".

```

Xayton (approved):

internal/orchestrator/bricks/bricks.go
martacarbone: This will change arduino-app-cli app list and the apps that appears in the inspitational (we want only kegacy exampels to appear here)

PR #546, opened 2026-07-27
check mtime of the linger file

giulio93:
### Motivation

Replace `loginctl show-user -p TimestampMonotonic `
with the `modtime` of systemd's own on-disk linger marker (/var/lib/systemd/linger/<user>). 
That file is only touched by enable-linger (created) / disable-linger (removed) and is untouched by a reboot, so its mtime is a real wall-clock fingerprint that stays valid across one — unlike the boot-relative monotonic value.

### DIFF
--- a/internal/orchestrator/pipewire/linger.go
+++ b/internal/orchestrator/pipewire/linger.go
@@ -112,14 +112,10 @@ func StopIfNotNeeded(ctx context.Context, cfg config.Configuration) error {
 		if err := disableLinger(ctx, user.Username); err != nil {
 			return fmt.Errorf("disable linger: %w", err)
 		}
-		err = clearOwner(ownerPath)
-		if err != nil {
-			return fmt.Errorf("clear linger ownership: %w", err)
-		}
 		slog.Debug("linger disabled", slog.String("username", user.Username))
 	}
 
-	return nil
+	return clearOwner(ownerPath)
 }
 
 func enableLinger(ctx context.Context, username string) error {
--- a/internal/orchestrator/pipewire/linger.go
+++ b/internal/orchestrator/pipewire/linger.go
@@ -13,28 +13,33 @@ import (
 	"os/user"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/arduino/go-paths-helper"
 
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/config"
 )
 
-func setOwner(ownerPath string, lingerTimestamp string) error {
+func setOwner(ownerPath string, lingerMtime time.Time) error {
 	if err := os.MkdirAll(filepath.Dir(ownerPath), 0700); err != nil {
 		return err
 	}
-	return os.WriteFile(ownerPath, []byte(lingerTimestamp), 0600)
+	return os.WriteFile(ownerPath, []byte(lingerMtime.Format(time.RFC3339Nano)), 0600)
 }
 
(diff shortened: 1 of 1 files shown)


internal/orchestrator/pipewire/linger.go
```diff
@@ -93,26 +95,31 @@ func StopIfNotNeeded(ctx context.Context, cfg config.Configuration) error {
...
 	}
-	if currentTimestamp == recordedTimestamp {
+
+	if currentMtime.Equal(recordedMtime) {
 		if err := disableLinger(ctx, user.Username); err != nil {
 			return fmt.Errorf("disable linger: %w", err)
 		}
+		err = clearOwner(ownerPath)
+		if err != nil {
+			return fmt.Errorf("clear linger ownership: %w", err)
```
Xayton: Keep style consistent with the rest of the file/function.
```suggestion
		if err = clearOwner(ownerPath); err != nil {
			return fmt.Errorf("clear linger ownership: %w", err)
```

internal/orchestrator/pipewire/linger.go
```diff
@@ -45,18 +50,15 @@ func clearOwner(ownerPath string) error {
...
-		return "", fmt.Errorf("create process %q: %w", strings.Join(cmdArgs, " "), err)
-	}
-
-	out, err := cmd.RunAndCaptureCombinedOutput(ctx)
+func lingerFileMtime(username string) (time.Time, error) {
+	info, err := os.Stat(filepath.Join("/var/lib/systemd/linger", username))
 	if err != nil {
-		return "", fmt.Errorf("%s: %w: %s", strings.Join(cmdArgs, " "), err, strings.TrimSpace(string(out)))
+		if os.IsNotExist(err) {
+			return time.Time{}, nil
```
Xayton, line 58: Why return this if the file does not exist? Wouldn't this hide an error?

internal/orchestrator/pipewire/linger.go
```diff
@@ -76,13 +78,13 @@ func runEnableLingerCmd(ctx context.Context, cfg config.Configuration) error {
...
 	if err != nil {
 		return fmt.Errorf("read linger state: %w", err)
 	}
 
 	ownerPath := cfg.DataDir().Join("pipewire").Join(fmt.Sprintf("linger-owner-%s", user.Username)).String()
-	return setOwner(ownerPath, timestamp)
+	return setOwner(ownerPath, mtime)
 }
 
 func StopIfNotNeeded(ctx context.Context, cfg config.Configuration) error {
```
Xayton, line 92: This function (`StopIfNotNeeded`) was previously removing the file (`clearOwner`) even if the timestamp does not match.

If we have the file, but the timestamp does not match, there is no reason to keep the owner file with a different timestamp.

If the timestamps don't match (someone else disabled + re-enabled linger between our two calls), our recorded mtime is stale forever and will never match again, yet the owner file is left behind as permanent garbage, and we lose the chance to reclaim ownership on a future `runEnableLingerCmd` (which currently doesn't rewrite the owner file when `alreadyOn == true`).

Xayton (changes requested):
Can we also add a unit test for this? I think it would help make sure some of these functions work correctly.

internal/orchestrator/pipewire/linger.go
giulio93: When this file does not exist the means that the linger is disable, so is the caller that need to handle the case.
if we have just enable it, it should exist, if we are disabling it should not

Xayton (approved):

PR #547, opened 2026-07-27
[Peripherals] Bluetooth support for Microphone / Speaker and generic D-Bus BT support 

rjtokenring:
### Motivation

# Bluetooth Audio Setup for UnoQ/VentunoQ (Speaker / Headset)

How to connect a Bluetooth audio device from the command line and make it visible to
PipeWire, so it can be used from an Arduino App via the `Speaker` and `Microphone`
peripherals.

Two boards are covered, because they behave differently:

| Board | OS | OS configuration required | Headset (mic) support |
|---|---|---|---|
| **Uno Q** | Debian | Yes — see [Uno Q](#uno-q-debian) | No (speaker only) |
| **Ventuno Q** | Ubuntu | None | Yes |

---

## Uno Q (Debian)

On Debian, a remote session (SSH) cannot see Bluetooth audio devices out of the box.
The configuration below is required.

> **Note:** on Uno Q only **speakers** (A2DP) can be used. Headset profiles (HSP/HFP,
> i.e. microphone) are not usable due to Bluetooth firmware limitations.

### 1. Disable seat monitoring for the BlueZ monitor

Without this, WirePlumber never starts its Bluetooth monitor in a session that has no
physical seat — which is the case for SSH.

```bash
sudo mkdir -p /etc/wireplumber/wireplumber.conf.d
sudo tee /etc/wireplumber/wireplumber.conf.d/80-bluez-headless.conf > /dev/null <<'EOF'
wireplumber.profiles = {
  main = {
    monitor.bluez.seat-monitoring = disabled
    monitor.bluez-midi.seat-monitoring = disabled
  }
}
EOF
```

### 2. Prevent the lightdm greeter from taking over the device

BlueZ hands the audio transport to a single media application. The lightdm greeter
starts its own PipeWire/WirePlumber stack at boot and claims the device before the
`arduino` user can. Masking the greeter's audio units avoids the conflict.

```bash
sudo mkdir -p /var/lib/lightdm/.config/systemd/user
for u in pipewire.service pipewire.socket wireplumber.service pipewire-pulse.service pipewire-pulse.socket; do
  sudo ln -sf /dev/null /var/lib/lightdm/.config/systemd/user/$u
done
sudo chown -R lightdm:lightdm /var/lib/lightdm/.config
```

Symlinking a unit to `/dev/null` masks it: systemd treats it as non-existent and will
not start it, not even through socket activation. Both `.service` and `.socket` units
must be masked, otherwise PipeWire is restarted on the first client request.

### 3. Enable linger for the `arduino` user

This starts the user audio stack at boot, without requiring a login session.

```bash
sudo loginctl enable-linger arduino # this is already done 
```

### 4. Reboot

```bash
sudo reboot
```

After the reboot, lightdm no longer claims the Bluetooth device and PipeWire in the
`arduino` session can see it.

### 5. Pair and connect the device

```bash
bluetoothctl scan on
bluetoothctl connect <MAC>
bluetoothctl trust <MAC>     # auto-reconnect on later boots
```

Verify:

```bash
wpctl status
```

The device should appear under **Audio → Devices** as `[bluez5]` and under
**Audio → Sinks**.

### 6. Use it from an Arduino App

```python
from arduino.app_peripherals.speaker import Speaker

speaker = Speaker(device="pipewire")
```

`device="pipewire"` uses the system default sink. To target a specific device, see
[Selecting a specific device](#selecting-a-specific-device).

---

## Ventuno Q (Ubuntu)

On Ubuntu everything works without any OS configuration.

### 1. Pair and connect the device

```bash
bluetoothctl scan on
bluetoothctl connect <MAC>
bluetoothctl trust <MAC>
```

The device appears in `wpctl status` right away.

### 2. Enable the headset profile (only if you need the microphone)

By default a headset is connected as a plain speaker, using **A2DP**. A2DP is
playback-only: it provides no microphone. To get both sink and source, switch to the
**HSP/HFP** headset profile.

List the available profiles:

```bash
pactl list cards | grep -A 40 -i <device-name>
```

Then select one, for example:

```bash
pactl set-card-profile bluez_card.00_02_5B_13_AE_86 headset-head-unit-msbc
```

| Profile | Sink | Source | Notes |
|---|---|---|---|
| `a2dp-sink` | yes | no | Stereo, SBC |
| `a2dp-sink-sbc_xq` | yes | no | Stereo, SBC-XQ (better quality) |
| `headset-head-unit-msbc` | yes | yes | Mono 16 kHz (HFP wideband) — **preferred for mic** |
| `headset-head-unit-cvsd` | yes | yes | Mono 8 kHz (narrowband) |

The two families are mutually exclusive: high-quality stereo playback and microphone
capture cannot be active at the same time. The selected profile is remembered across
reconnections.

> **Tip:** WirePlumber's `bluetooth.autoswitch-to-headset-profile` setting (enabled by
> default) switches to HFP automatically when an application opens a capture stream, and
> back to A2DP when it closes. Leaving A2DP as the resting profile is often enough.

### 3. Use it from an Arduino App

```python
from arduino.app_peripherals.speaker import Speaker
from arduino.app_peripherals.microphone import Microphone

speaker = Speaker(device="pipewire")
mic = Microphone(device="pipewire")
```

---

## Setting the default device

`device="pipewire"` follows whatever the user has selected as default, which is usually
the friendliest option: the application follows the system, and no node names need to be
hardcoded.

```bash
wpctl status                 # the '*' marks the current default
wpctl set-default <ID>       # works for both sinks and sources
```

By name, which is more stable in scripts:

```bash
pactl set-default-sink   bluez_output.00_02_5B_13_AE_86.1
pactl set-default-source bluez_input.00_02_5B_13_AE_86.0
```

---

## Selecting a specific device

```python
speaker = Speaker(device="pipewire:NODE=<node.name>")
mic = Microphone(device="pipewire:NODE=<node.name>")
```

Note the syntax: `NODE=<value>`, with `=`.

### Finding node names

The name shown by `wpctl status` is the human-readable *description*
(e.g. `TaoTronics TT-BH026`), **not** the `node.name` expected by the ALSA string. Get
the actual names with:

```bash
pactl list sinks short
pactl list sources short
```

The second column is the value to use. BlueZ nodes follow the pattern
`bluez_output.<MAC>.<n>` and `bluez_input.<MAC>.<n>`, with underscores instead of colons
in the MAC — no spaces, so no quoting is needed.

`pw-dump` and `wpctl inspect <ID>` also expose `node.name`, with more detail.

Two caveats:

- Do **not** use the numeric ID from `wpctl status`. A number in the ALSA string is
  interpreted as `object.serial`, which is a different value and changes on every
  reconnection.
- The numeric suffix of the node name depends on the active profile, so it changes when
  switching between A2DP and HFP.
- A `.monitor` source (e.g. `bluez_output.<MAC>.1.monitor`) is not a microphone: it is a
  loopback of what is being played back.

### Resolving the name at runtime

Since node names are not stable across profile changes, resolving by MAC is more robust
than hardcoding:

```python
import subprocess

def resolve_node(kind: str, mac: str) -> str | None:
    """kind: 'sinks' or 'sources'."""
    out = subprocess.check_output(["pactl", "list", kind, "short"], text=True)
    tag = mac.replace(":", "_")
    for line in out.splitlines():
        name = line.split("\t")[1]
        if tag in name and not name.endswith(".monitor"):
            return name
    return None

node = resolve_node("sinks", "00:02:5B:13:AE:86")
speaker = Speaker(device=f"pipewire:NODE={node}")
```

`device="pipewire"` and `device="pipewire:NODE=-1"` are equivalent: `-1` is the default
value of the `NODE` argument and means "let the session manager decide". Prefer the plain
`"pipewire"` form unless the string is built programmatically.

# D-Bus support

This PR also add BT support to main python container through D-Bus.
Tested with:

```python
busctl --system tree org.bluez        # se questo funziona, sei a posto
python -c "import asyncio,bleak; print(asyncio.run(bleak.BleakScanner.discover()))"
```

### Reviewer checklist


internal/orchestrator/peripherals/devices.go
```diff
@@ -69,6 +86,76 @@ func Detect(ctx context.Context) (AvailableDevices, error) {
...
+}
+
+// HasBluetoothMicrophone reports whether a bluetooth microphone (a bluez5
+// Audio/Source, as exposed by a headset) is currently connected, by inspecting
+// the PipeWire graph via pw-dump. If pw-dump is not available it returns false.
+func HasBluetoothMicrophone(ctx context.Context) bool {
+	nodes, ok := pwDumpNodes(ctx)
+	return ok && hasBluetoothAudioNode(nodes, audioSourceClass)
+}
+
```
lucarin91: Those don't seem to be used

lucarin91 (commented):
I would add also the configuration for making BT audio work with unoq in the `arduino-unoq-config`

rjtokenring:
> I would add also the configuration for making BT audio work with unoq in the `arduino-unoq-config`

all required steps for fixing it. I just reported how to make it work: then it should be added to build and to the process that are in charge of deploying conf changes during auto update.

internal/orchestrator/peripherals/devices.go
rjtokenring: removed

PR #548, opened 2026-07-28
Gate device detection and pipewire on app requirements

mirkoCrobu:
### Change description

- detect peripherals only when the app requires them
- start PipeWire only for apps that require audio

### Reviewer checklist


Xayton (commented):
Changes are good.
I would change the returned list of required device classes from a map to a slice.
You can continue using a map in `requiredDeviceClasses` to ensure uniqueness of the keys.

Xayton (approved):

PR #557, opened 2026-07-30
feat: set media-carrier built-in mic to base volume on first install

### LINKED ISSUE #549: Set volume for the media-carrier built-in microphone
Pipewire start media-carrier microphone to 100% so the recording audio is just noise.
We need a way to set the mic at the right level in order to be usable.

giulio93:
## Summary
- Pre-seeds WirePlumber's own state file (`~/.local/state/wireplumber/default-routes`) with the on-board microphone's default volume (22%) directly during `postinst`, gated by a flag file to run exactly once.
- WirePlumber matches restore entries by route key whenever the route is (re)created, regardless of whether it existed yet when the value was written — verified empirically on hardware.
- Subsequent user/app changes to the volume are left untouched on reinstall/upgrade

## Test plan
- [x] Simulated fresh install (flag + existing state entry removed), rebuilt and installed the `.deb`, confirmed `postinst` correctly appended the entry to the existing file.
- [x] Restarted `wireplumber.service` and confirmed `wpctl` reports the mic at 22%.
- [x] Confirmed idempotency: manually changed volume to 60%, reinstalled the package again, confirmed the value was **not** reset back to 22%.

### DIFF
--- a/debian/arduino-app-cli/DEBIAN/postinst
+++ b/debian/arduino-app-cli/DEBIAN/postinst
@@ -17,3 +17,4 @@ fi
 
 #LOAD_HELPER
 configure_agent_profiles
+configure_mic_default_volume
--- a/debian/arduino-app-cli/DEBIAN/postinst
+++ b/debian/arduino-app-cli/DEBIAN/postinst
@@ -17,4 +17,4 @@ fi
 
 #LOAD_HELPER
 configure_agent_profiles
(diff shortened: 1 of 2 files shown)


Xayton:
Superseded by https://github.com/bcmi-labs/arduino-deb-packages/pull/98 (same change, but done in the `arduino-unoq-config` package).

PR #564, opened 2026-08-04
Enable pipewire via adb shell

giulio93:
### Motivation
In order to let PW be able to start up in USB mode for VENTUNO Q, we need to add it to the start up cases


Xayton (changes requested):
I would try to simplify the problem:
lets start Pipewire (and configure lingering) every time we start an app that requires audio (`needsAudioDevices`). For all cases and all boards (both UNO Q and VENTUNO Q).

I think it should work. But needs to be tested.

Xayton (approved):

PR #567, opened 2026-08-04
Filter out unsupported bricks

martacarbone:
### Motivation

Refactored the examples.Bricks processing logic in GetExamples to filter out unsupported bricks 

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/examples.go
+++ b/internal/orchestrator/examples.go
@@ -57,19 +57,18 @@ func GetExamples(cfg config.Configuration, brickIndex *bricksindex.BricksIndex, 
 		retrieveExamplesInfo(examples.CoreAndFoundational[i].Examples, idProvider)
 	}
 
-	for i := range examples.Bricks {
-		addBrickInfo(&examples.Bricks[i], brickIndex)
-		retrieveExamplesInfo(examples.Bricks[i].Examples, idProvider)
+	rawBricks := examples.Bricks
+	examples.Bricks = make([]BrickExamples, 0, len(rawBricks))
+	for i := range rawBricks {
+		brickData, found := brickIndex.FindBrickByID(rawBricks[i].Brick)
+		if found {
+			rawBricks[i].BrickCategory = brickData.Category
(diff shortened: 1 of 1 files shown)


internal/orchestrator/examples.go
```diff
@@ -57,21 +57,20 @@ func GetExamples(cfg config.Configuration, brickIndex *bricksindex.BricksIndex,
...
-		addBrickInfo(&examples.Bricks[i], brickIndex)
-		retrieveExamplesInfo(examples.Bricks[i].Examples, idProvider)
+	rawBricks := examples.Bricks
+	examples.Bricks = make([]BrickExamples, 0, len(rawBricks))
+	for i := range rawBricks {
+		brickData, found := brickIndex.FindBrickByID(rawBricks[i].Brick)
+		if found {
+			rawBricks[i].BrickCategory = brickData.Category
+			retrieveExamplesInfo(rawBricks[i].Examples, idProvider)
+			examples.Bricks = append(examples.Bricks, rawBricks[i])
```
dido18, line 67: Is it on purpose to silently drops missing bricks ?
we could log a debug

```suggestion
		if !found {
		    slog.Debug("Brick referenced in examples not found in index", slog.String("brick",    rwBricks[i].Brick))
    continue
		}
		rawBricks[i].BrickCategory = brickData.Category
		retrieveExamplesInfo(rawBricks[i].Examples, idProvider)
		examples.Bricks = append(examples.Bricks, rawBricks[i])
```

dido18 (commented):
I suggest adding unit tests to avoid regression

mirkoCrobu (approved):

mirkoCrobu (commented):
agree with @dido18

internal/orchestrator/examples.go
martacarbone: This is intentional; I didn't add logging because it would be too frequent.

martacarbone: A filtering test added in this PR: https://github.com/arduino/arduino-app-cli/pull/570

PR #571, opened 2026-08-05
Use system init data in upgrade progress info

mirkoCrobu:
### Motivation

- Progress bar froze at 44% during the Docker image download
- That is the longest step of the update: no feedback for minutes

### Change description

- `system init` is now called with `--format json-lines`
- Its JSON payload moved to `internal/orchestrator` (`InitResult`): shared contract between the two processes(init and update)
- The update process parses each line back into `InitResult`
- Only `source: docker` progress is used
- `source: arduino` progress is per file on an unknown total, so it is only logged
- The 0-100 percent of `system init` is mapped into the local 30-90 band of `system update`, and clamped
- Progress never goes back: in case of init failure, it will restart from the last percentage reached
- Lines that are not valid JSON (stderr) fall back to log lines
- Tests on the line parser and on the band mapping

### Additional Notes

tested manually on UnoQ

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/init_result.go
+++ b/internal/orchestrator/init_result.go
@@ -13,14 +13,6 @@ const (
 	InitResultProgress = "progress"
 )
 
-// InitResult is the serializable form of an InitEvent: it is what the
-// `system init` command writes, one JSON object per line, when asked for the
-// json-lines output format.
-//
-// It is a contract, not just a rendering: the update process runs `system init`
-// as a subprocess and reads these lines back to follow the progress of the
-// docker images download, so the field names are part of an interface between
-// the two.
 type InitResult struct {
 	Type    string `json:"type"`
 	Source  string `json:"source,omitempty"`
--- a/internal/orchestrator/init_result.go
+++ b/internal/orchestrator/init_result.go
@@ -15,8 +15,8 @@ const (
 )
 
 type InitResult struct {
-	Type    InitResultType `json:"type"`
-	Source  string         `json:"source,omitempty"`
+	Type    InitResultType  `json:"type"`
+	Source  InitEventSource `json:"source,omitempty"`
 	Message string         `json:"message,omitempty"`
 	Label   string         `json:"label,omitempty"`
 	Current int64          `json:"current,omitempty"`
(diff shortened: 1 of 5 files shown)


internal/orchestrator/init_result.go
```diff
@@ -0,0 +1,69 @@
...
+import "fmt"
+
+// Types of an InitResult.
+const (
+	InitResultLog      = "log"
+	InitResultProgress = "progress"
+)
+
+type InitResult struct {
+	Type    string `json:"type"`
```
cmaglie: Can we create an enumeration type for `Type`? something like

```suggestion
type InitResultType string

// Types of an InitResult.
const (
	InitResultLog      InitResultType = "log"
	InitResultProgress InitResultType = "progress"
)

type InitResult struct {
	Type    InitResultType `json:"type"`
```

It would help language servers to give meaningful suggestions.
I don't know if the json encoder may digest it automatically tho, should be tested.

internal/update/apt/service.go
```diff
@@ -125,7 +126,23 @@ func (s *Service) UpgradePackages(ctx context.Context, packages []update.Package
 		eventCB(update.NewDataEvent(update.UpgradeLineEvent, line))
 	}
 	eventCB(update.NewProgressEvent("docker images download", imagesDownloadProgress))
-	for line, err := range runSystemInit(ctx) {
+
+	// Of the two sources reporting a progress, only docker can be trusted: it is
+	// the fraction of the bytes of every image, and the total is known before the
+	// download starts. The arduino one is per file, out of an unknown number of
+	// them, so it is only logged.
```
lucarin91, line 133: So why are we returning that progress? I am still confused about what we are trying to achieve here. If we don't have reliable progress for the whole system init operation, why are we reporting progress? I think the new improved logs are enough anyway.

cmaglie: I think the reason is summarized in the issue:

> Motivation
> * Progress bar froze at 44% during the Docker image download
> * That is the longest step of the update: no feedback for minutes

Maybe a finer progress in one step of the multi-step (the longest one in this case), is far better than no finer progress in any of the step.

mirkoCrobu: Docker image pull is the longest step of the update, taking several minutes. Without parsing it, the bar just jumps from 30 to 90 and then stays there. Parsing it gives progress that moves gradually instead of one fixed jump.
We only use source: docker because it's the reliable one. Arduino total is unknown a priori, so we only log it.

cmaglie (approved):

Issue #184, opened 2026-01-07
Add basic brick functionality to the CLI


Xayton:
Currently the arduino-app-cli, when used as a CLI, supports only:
- listing bricks
- showing details of a given brick

Add some other functions (exposed by [the service](https://github.com/arduino/arduino-app-cli/blob/main/internal/orchestrator/bricks/bricks.go)), that are already available in the APIs, like:
- list the bricks of a given app
- add and remove a brick from an app

lucarin91:
In more detail, I think we need to implement the following commands in the app-cli to be in line with the api endpoints:

1. Add a brick to an app
```
app brick add <brick-id> [--app <id|path>] [--model <model-id>] [--var KEY=VALUE]...
```

2. Remove a brick from an app
```
app brick remove <brick-id> [--app <id|path>]
```

3. get configuration of the brick including variable and model (if present)
```
app brick get-config <brick-id> [--app <id|path>]
```
```
MODEL yolox-object-detection

VARIABLE               VALUE                                             REQUIRED
EI_OBJ_DETECTION_MODEL /home/arduino/.arduino-bricks/models/face-det.eim true
DETECTION_THRESHOLD    0.5 (default)                                     false
```

4. Set/update configuration of a brick
```
app brick set-config <brick-id> [--app <id|path>] [--model <model-id>] [--var KEY=VALUE]...
```

5. List all brick of an app
```
app brick list [--app <id|path>]
```
```
ID                       NAME
arduino:object_detection Object Detection yolox-object-detection
arduino:video_stream     Video Stream
my_local_brick           My Local Bric
```

In the commands the `--app` is optional and, by default, is the current folder; if the current folder isn't an app, the command will return an error.

PR #575, opened 2026-08-08
Restore --bricks support in app new

### LINKED ISSUE #568: `--bricks` flag of `app new` command is non-functional
### Describe the problem

The `app new` command accepts a `--bricks` flag:

```text
$ arduino-app-cli app new --help
Creates a new Arduino App

Usage:
  arduino-app-cli app new name [flags]

Flags:
  -b, --bricks stringArray   List of bricks to include in the app
```

🐛 Contrary to what the command line reference promises, Arduino App CLI completely ignores the list of Bricks the user passes as an argument to this flag.

### To reproduce

```text
$ arduino-app-cli app new --bricks "arduino:air_quality_monitoring" "foo"

App created successfully: /home/arduino/ArduinoApps/foo (ok)

$ cat ~/ArduinoApps/foo/app.yaml 

# app.yaml: The main configuration file for your Arduino App.
# This file describes the application's metadata and properties.

# The user-visible name of the application.
name: foo

# The icon for the application, can be an emoji or a short string.
icon:

# A list of network ports that the application exposes.
# Example: [80, 443]
ports: []

# A list of bricks used by this application.
bricks: []
```

🐛 The `arduino:air_quality_monitoring` brick specified by the `app new` command was not added to the `bricks` sequence in the `user:foo` App metadata.

### Expected behavior

The newly configured App is configured with the list of Bricks passed to the `--bricks` flag (https://github.com/arduino/arduino-app-cli/issues/184).

**-OR-**

The `app new` command does not have a non-functional `--bricks` flag.

### Arduino App CLI version

a7a1b3d

### Additional context

The regression occurred at https://github.com/arduino/arduino-app-cli/commit/328b969d2465e3a755a08549d5d3e1a0b76aa802, which removed the code for handling the flag but did not also remove the flag.

---

Originally reported at https://forum.arduino.cc/t/uno-q-creating-a-new-app-including-bricks-on-uno-q-using-arduino-app-cli/1454309.

### Issue checklist

Daniiiil1:
## Motivation

The `app new --bricks` flag is still exposed but its values are dropped before app generation, so every generated `app.yaml` contains an empty `bricks` list.

## Change description

- thread the CLI brick IDs through `createHandler` and `CreateAppRequest`
- populate the generated `AppDescriptor` with those brick IDs
- render brick IDs in `app.yaml` while preserving `bricks: []` for the default case
- add a round-trip regression test for multiple brick IDs

The IDs are quoted in the template so unusual but valid values remain valid YAML.

## Testing

- `go test ./internal/orchestrator/app/generator`
- `go test ./internal/orchestrator -run 'TestCreateApp|Test.*App' -count=1`

Closes #568

### DIFF
--- a/internal/orchestrator/app/generator/app_template/app.yaml.template
+++ b/internal/orchestrator/app/generator/app_template/app.yaml.template
@@ -12,4 +12,11 @@ icon: {{ .Icon }}
 ports: [{{ joinInts .Ports }}]
 
 # A list of bricks used by this application.
+{{- if .Bricks }}
+bricks:
+{{- range .Bricks }}
+  - {{ printf "%q" .ID }}
+{{- end }}
+{{- else }}
 bricks: []
+{{- end }}
--- a/cmd/arduino-app-cli/app/new.go
+++ b/cmd/arduino-app-cli/app/new.go
@@ -84,7 +84,7 @@ func createHandler(
 			Description: description,
 			Bricks:      bricks,
 			SkipSketch:  noSketch,
-		}, servicelocator.GetAppIDProvider(), cfg)
+		}, servicelocator.GetBricksIndex(), servicelocator.GetAppIDProvider(), cfg)
 		if err != nil {
 			feedback.Fatal(err.Error(), feedback.ErrGeneric)
 			return nil
--- a/cmd/arduino-app-cli/app/new.go
+++ b/cmd/arduino-app-cli/app/new.go
@@ -32,7 +32,7 @@ func newCreateCmd(cfg config.Configuration) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cobra.MinimumNArgs(1)
 			name := args[0]
-			return createHandler(cfg, name, icon, description, noSketch, fromApp)
+			return createHandler(cfg, name, icon, description, bricks, noSketch, fromApp)
 		},
 	}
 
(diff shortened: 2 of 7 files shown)


internal/orchestrator/app/generator/app_template/app.yaml.template
```diff
@@ -12,4 +12,11 @@ icon: {{ .Icon }}
 ports: [{{ joinInts .Ports }}]
 
 # A list of bricks used by this application.
+{{- if .Bricks }}
+bricks:
+{{- range .Bricks }}
+  - {{ printf "%q" .ID }}
+{{- end }}
+{{- else }}
```
lucarin91, line 20: I would avoid adding bricks with a template and instead use this method https://github.com/arduino/arduino-app-cli/blob/7fe73948b5bbfdb5cf19bb07c2d79782dddc72ea/internal/orchestrator/bricks/bricks.go#L341, which also checks if the brick exists and if it requires some variables.

lucarin91 (changes requested):
Thanks for the contribution; we probably forgot about that flag during development 😄 . Not sure if we should keep that or not, but if we want to implement it I would do as suggested in the comment below.

per1234:
> Not sure if we should keep that or not

@lucarin91 there is an indication that it is desired here: https://github.com/arduino/arduino-app-cli/issues/184

Note the commit that caused the regression (https://github.com/arduino/arduino-app-cli/commit/328b969d2465e3a755a08549d5d3e1a0b76aa802) was to address `bcmi-labs/orchestrator#464`. So the stated goal of that change was to remove redundancy from the daemon API. From this, my interpretation is that the change's breakage of the `--bricks` CLI flag was unintentional. The correct change would have been to reimplement the flag's functionality using the same infrastructure used by the `PUT /v1/apps/{appID}/bricks/{brickID}` API endpoint. But that wasn't done.

lucarin91 (approved):
I think it is ok too keep the brick creation via template, but I added a check to be sure that the brick name passed by the user exists in our index.

Issue #523, opened 2026-07-10
Improve error reporting when downloading an ai model fails


Xayton:
For example, in the case of [this issue](https://github.com/bcmi-labs/arduino-ubuntu-images/pull/103) (/var/lib/arduino-app-cli/models owned by root) no clear error is returned or logged.

PR #583, opened 2026-08-13
fix(dockerhelper): fail when a bind mount is not usable by the container

giulio93:
> **Stacked on #581** — base is `fix/model-download-error-reporting`, so only the last commit belongs to this PR. Retarget to `main` once #581 merges.

Follow-up to #523. #581 makes a failed download report *what* went wrong; this stops the daemon causing one class of failure at all.

## The problem

`Run` pre-creates each bind mount's host directory and swallowed the error when it could not:

```go
if err := os.MkdirAll(hostPath, 0775); err != nil {
    slog.Warn("cannot pre-create bind mount directory", "path", hostPath, "err", err)
    continue                      // ← proceeds into a known-broken state
}
```

That pre-creation is the only thing preventing Docker from creating the path, and Docker runs as root. So when it fails:

1. the daemon (uid 1000) cannot create `<models>/<repo>` → `EACCES` → warned at a level the shipped unit filters out, then ignored
2. Docker creates the bind path itself, **root-owned `755`**
3. the container runs as the invoking user and cannot write to its own `/models`
4. the script dies with `PermissionError: [Errno 13] Permission denied: '/models/unsloth'`

The user is shown a path *inside the container* for a host problem, and step 2 leaves a root-owned directory that keeps failing until someone chowns it.

## The change

`ensureBindDirs` creates the directory, verifies it is writable by the user the container runs as, and returns an error naming the path, its owner and its mode. Read-only mounts only need to be readable.

`HandleInstallModel` now forwards the error from `GetModelByID` instead of replacing it with the fixed string `"unable to get model by ID"` — otherwise the diagnostic never reaches the caller. `HandlerModelByID` already forwarded `err.Error()`; the two were inconsistent.

## A stale fixture this uncovered

The new check failed three existing tests, for a real reason. `testdata/models-handlers.yaml` used `${ARDUINO_APP_BRICKS__CUSTOM_MODEL_DIR}`, which `loadHandlers` never puts in `configEnv` — only `DOCKER_REGISTRY_BASE`, `BOARD_NAME` and `MODELS_PATH`. Unset variables substitute to the empty string, so every bind in those tests resolved to `/`. They had been mounting the filesystem root. The same tests passed `paths.New("testdata/models")`, a directory that does not exist in the repo.

The fixture now uses `${MODELS_PATH}` as the shipped handlers do, and those tests get a real `t.TempDir()`.

Separately: `ResolveVars`' comment claims unknown variables are left unchanged, but compose-go substitutes `""` and returns no error, so any typo in a handler's `volumes` silently produces a wrong bind. Not addressed here.

## Testing

`dockerhelper_test.go` covers create-missing, exists-writable, cannot-create, exists-unwritable, read-only-unwritable and malformed binds, plus `:ro` parsing; permission cases skip under root. Validated on a UNO Q — evidence below.

`go build`, `go vet`, `gofmt` clean; `go test ./internal/...` passes except the pre-existing Docker-dependent `TestRemoveImage`.

## For reviewers

This makes a previously silent condition fatal. A board whose `/var/lib/arduino-app-cli/models` is already root-owned — the state the old code kept producing — will now fail loudly on listing and install instead of obscurely later. Intended, but a conscious call. The `chown` in `DEBIAN/postinst` only runs at install time, so an affected board needs `chown -R arduino:arduino /var/lib/arduino-app-cli` or a reinstall.

### DIFF
--- a/internal/api/handlers/ai_models.go
+++ b/internal/api/handlers/ai_models.go
@@ -179,8 +179,10 @@ func HandleInstallModel(dockerClient command.Cli, modelsIndex *modelsindex.Model
 
 		model, err := modelsIndex.GetModelByID(r.Context(), id)
 		if err != nil {
+			// Forward the reason, as HandlerModelByID does: it names the model
+			// and the path that could not be used, which the caller needs.
 			slog.Error("unable to get model by ID", slog.String("error", err.Error()))
-			render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "unable to get model by ID"})
+			render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: err.Error()})
 			return
 		}
 		if model == nil {
--- a/internal/dockerhelper/dockerhelper.go
+++ b/internal/dockerhelper/dockerhelper.go
@@ -16,8 +16,11 @@ import (
 	"log/slog"
 	"os"
 	"os/user"
+	"path/filepath"
+	"slices"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/docker/docker/api/types/container"
@@ -54,12 +57,8 @@ func Run(ctx context.Context, cli client.APIClient, opts RunOptions) error {
 	stderrTail := &tailBuffer{}
 	stderr := io.MultiWriter(opts.Stderr, stderrTail)
 
-	for _, bind := range opts.Binds {
-		hostPath, _, _ := strings.Cut(bind, ":")
-		if err := os.MkdirAll(hostPath, 0775); err != nil {
-			slog.Warn("cannot pre-create bind mount directory", "path", hostPath, "err", err)
-			continue
-		}
+	if err := ensureBindDirs(opts.Binds); err != nil {
+		return err
 	}
 
 	launchStart := time.Now()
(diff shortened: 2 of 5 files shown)


giulio93:
Hardware validation, UNO Q, `--log-level error` as shipped.

**Before** — the daemon swallowed the error and let Docker create the path as root:

```
WARN cannot pre-create bind mount directory path=/tmp/mt/models/llamacpp
     err="mkdir /tmp/mt/models/llamacpp: permission denied"

drwxr-xr-x 2 root root 40 /tmp/mt/models/llamacpp     ← created by Docker
```

and the client saw only the generic message:

```json
{"details":"unable to get model by ID"}
```

**After** — same unwritable models directory:

```json
{"details":"cannot determine install status for model \"llamacpp:Qwen3.5-0.8B-Q4_0\": model check failed for \"llamacpp:Qwen3.5-0.8B-Q4_0\": cannot create bind mount directory \"/tmp/mt/models/llamacpp\" (owned by 1000:1000, mode 0555): mkdir /tmp/mt/models/llamacpp: permission denied"}
```

```
$ ls -la /tmp/mt/models/
total 0
dr-xr-xr-x 2 arduino arduino 40 .
drwxrwxr-x 3 arduino arduino 60 ..
```

No root-owned directory is left behind. The same text also reaches the journal at `ERROR` level, so it survives the shipped log level.

**Healthy path unaffected** — writable models directory on the same build: 13 models listed, and a real GGUF download ran to 70% before I cut it.

Method: run as `arduino` on a spare port with `MODELS_PATH` pointed at a scratch directory, so the packaged systemd deployment was not exercised.

PR #584, opened 2026-08-13
Add board id detection

martacarbone:
### Motivation

Add board id identification.
Required for arduino-linux-config.

### Reviewer checklist


lucarin91 (changes requested):
The `board` library is ment to be used also remotelly to connect to a board. This is a function isntead that require to read the device tree of the board, so it work only locally. The `devicetree` library already give a good abstraction for reconis a board, I would avoid adding another function.

martacarbone:
Use the device tree library in the arduino-linux-code.

PR #589, opened 2026-08-14
feat: implement build command

lucarin91:
### Change description

Adds `arduino-app-cli app build <app_path>`, which builds an app into a `<name>-<version>-<board>.arduinoapp` archive (gzipped tar, so the venv keeps its symlinks and exec bits). It holds `release.yaml` (the manifest: schema, name, version, target, notes and the bricks, models and libraries the app was built with, written first so a reader needs nothing else), `src/` (the app as authored, with `app.yaml` redacted of brick secrets) and `prebuild/`, which becomes `.cache/` on install: the frozen compose files, the indexes cropped to what the app uses, and the prebuilt python environment.

Flags: `--target`, `--version`, `--notes` (file or `-`), `--output`/`-o`, `--overwrite`.

### Additional Notes

The install side is not in this PR: nothing reads a `.arduinoapp` archive yet.


internal/orchestrator/release.go
```diff
@@ -0,0 +1,554 @@
...
+	file, err := archivePath.Create()
+	if err != nil {
+		return fmt.Errorf("failed to create %s: %w", archivePath, err)
+	}
+	defer file.Close()
+
+	gzipWriter := gzip.NewWriter(file)
+	defer gzipWriter.Close()
+	tarWriter := tar.NewWriter(gzipWriter)
+	defer tarWriter.Close()
```
Xayton: file/gzipWriter/tarWriter are defer closed and any error while closing is dropped. If any of the close fails, the corresponding archive could be reported as OK while being truncated (missing the gzip or tar footer).

Better a defer that sets the function error and closes them in sequence, while checking for errors:
```go
defer func() {
   for _, closer := range []io.Closer{tarWriter, gzipWriter, file} {
       if closeErr := closer.Close(); closeErr != nil && err == nil {
           err = closeErr
       }
   }
   if err != nil {
       _ = archivePath.Remove()
       err = fmt.Errorf("failed to write %s: %w", archivePath, err)
   }
}()
```

Also, this way the archivePath.Remove() (when needed) always runs after the Close operations.

internal/orchestrator/release.go
```diff
@@ -0,0 +1,554 @@
...
+	}
+
+	// TODO: fail when a service a brick requires is missing for the target. The
+	// generator skips it, so it would be missing from the archive for good.
+
+	version := req.Version
+	if version == "" {
+		version = time.Now().UTC().Format("20060102-150405")
+	}
+	name := slug.Make(appToBuild.Name)
```
Xayton: Low priority: version is not escaped. So a malicious version would escape the output directory.
For example: `--version ../../evil`
```suggestion
	version = slug.Make(version)
	name := slug.Make(appToBuild.Name)
```

internal/orchestrator/release.go
```diff
@@ -0,0 +1,554 @@
...
+	if err != nil {
+		return BuildReleaseResult{}, fmt.Errorf("the staged app is not valid: %w", err)
+	}
+
+	// Before the environment, as in a start: it is quick and it is where a build
+	// fails, so it does not make anyone wait for the venv to find it out.
+	cb(StreamMessage{data: "freezing the compose files", progress: &Progress{Name: "compose files", Progress: 10.0}})
+	appEnv := appEnvironment(ctx, stagedApp, bricksIndex, modelsIndex, plat)
+	buildOpts := BuildOptions{
+		ProjectName: slug.Make(releaseName),
```
Xayton: ProjectName contains the app release version.
This means that many of the resources created by docker (like containers, networks, etc.) when running the compose, will contain the version as well.

This could create issues when upgrading a release, as the new version would not overwrite the previous one. We'd need to stop and clean up everything docker created manually.

I suggest to, at minimum, remove the version from the compose project name.
Even better: move the project name generation from resolve to render, so that it can take into account also the path where the app release is installed, as a way to uniquely identify it when it's running in docker.

Xayton (approved):

PR #591, opened 2026-08-18
feat: emit error on port collisions when starting an app

mirkoCrobu:
### Motivation
Two bricks can ask for the same host port. If they run inside `main`, Compose merges
the duplicate, and the app starts with one UI silently unreachable. If they run in their
own containers, Docker refuses to bind, but only after provisioning and image pull, and
the error names the port, never the bricks.

### Change description

1. `checkPortCollisions` collects the ports of every source of the app and fails the
   start when one is declared more than once. . Errors are joined:

   ```
   port 7000 is declared by multiple sources: arduino:web_ui, arduino:streamlit_ui
   ```

   Sources: `app.yaml`, the brick index, the `brick_compose.yaml` files, and the
   `service_compose.yaml` of the required services. Services are reported against the
   brick requiring them, and count as one source even when required by several bricks.

2. Compose port extraction moved to `internal/composefile`, so `servicesindex` can reuse
   it. `servicesindex.Service` gains `containerPorts` + `GetPorts`, like `bricksindex.Brick`.

3. Unit tests per source, plus two e2e tests for the two cases above.

### Reviewer checklist

### DIFF
--- a/internal/orchestrator/check.go
+++ b/internal/orchestrator/check.go
@@ -13,6 +13,7 @@ import (
 	"log/slog"
 	"maps"
 	"slices"
+	"strconv"
 
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/app"
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/bricksindex"
@@ -69,6 +70,48 @@ func checkBricks(ctx context.Context, bricks []app.Brick, index *bricksindex.Bri
 	}
 
 	return allErrors
+}
+
+// appPortsSource is the collision source name used for the ports declared in the app.yaml file.
+const appPortsSource = "app.yaml"
+
+// portCollision is a port declared by more than one source.
+type portCollision struct {
+	Port    string
+	Sources []string
+}
+
+func detectPortCollisions(appPorts []int, bricks []app.Brick, index *bricksindex.BricksIndex) []portCollision {
+	sourcesByPort := make(map[string][]string)
+	addSource := func(port, source string) {
+		if !slices.Contains(sourcesByPort[port], source) {
+			sourcesByPort[port] = append(sourcesByPort[port], source)
+		}
+	}
+
+	for _, p := range appPorts {
+		addSource(strconv.Itoa(p), appPortsSource)
+	}
+
+	for _, appBrick := range bricks {
+		indexBrick, found := index.FindBrickByID(appBrick.ID)
+		if !found {
+			continue
+		}
+
+		for _, p := range indexBrick.GetPorts() {
+			addSource(p, appBrick.ID)
+		}
+	}
+
+	var collisions []portCollision
+	for _, p := range slices.Sorted(maps.Keys(sourcesByPort)) {
+		if len(sourcesByPort[p]) > 1 {
+			collisions = append(collisions, portCollision{Port: p, Sources: sourcesByPort[p]})
+		}
+	}
+
+	return collisions
 }
 
 // requiredDeviceClasses returns the sorted device classes required by the app bricks, skipping the
--- a/internal/orchestrator/check.go
+++ b/internal/orchestrator/check.go
@@ -14,6 +14,7 @@ import (
 	"maps"
 	"slices"
 	"strconv"
+	"strings"
 
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/app"
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/bricksindex"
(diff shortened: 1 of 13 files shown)


internal/orchestrator/orchestrator.go
```diff
@@ -113,6 +113,16 @@ func StartApp(
...
 		return err
 	}
+
+	for _, collision := range detectPortCollisions(appToStart.Descriptor.Ports, appToStart.Descriptor.Bricks, bricksIndex) {
+		slog.Warn("port collision detected", slog.String("port", collision.Port), slog.Any("sources", collision.Sources))
+		// TODO convert this eventn in warning type after https://github.com/arduino/arduino-app-cli/issues/382 is approved and handled by App Lab
+		cb(StreamMessage{data: fmt.Sprintf(
+			"[WARNING] port %s is declared by more than one source (%s): only one of them will be reachable",
+			collision.Port, strings.Join(collision.Sources, ", "),
+		)})
```
lucarin91: I would return an error and refuse to start the app

mirkoCrobu:
### test 1 
both bricks expose port 7000
```yaml
name: "Detect objects on images Copy of "
description: Object detection in the browser
ports: []
bricks:
- arduino:web_ui:
- arduino:object_detection: {}
- arduino:streamlit_ui: {}
icon: 🏞️
```
image

mirkoCrobu:
### test 2 
app,yaml exposes the same port used by a service

app.yaml
```yaml
name: test services port collision
description: ""
ports:
- 9999
bricks:
- arduino:llm:
    model: llamacpp:gemma-3-1b-it-Q4_0
icon: 👀
```
[brick LLM](https://github.com/arduino/app-bricks-py/blob/main/src/arduino/app_bricks/llm/brick_config.yaml) requires a [service that exposes port 999](https://github.com/arduino/app-bricks-py/blob/main/src/arduino/app_services/llamacpp/service_compose.yaml)

#### result

image

internal/orchestrator/provision.go
lucarin91, line 1: I will revert the change on this file, those aren't necessary for your PR

internal/orchestrator/check.go
```diff
@@ -71,6 +74,120 @@ func checkBricks(ctx context.Context, bricks []app.Brick, index *bricksindex.Bri
...
+				slog.Debug("service required by brick not found or not available for current board", slog.String("service_id", id), slog.String("brick_id", appBrick.ID))
+				continue
+			}
+			services[id] = *service
+		}
+	}
+
+	return services, nil
+}
+
```
lucarin91: I would just use a single function here. Note also that for now service concept isn't expose to the user, so I would return the brick id and not the service one

```suggestion
func checkPortCollisions(
	app app.AppDescriptor,
	index *bricksindex.BricksIndex,
	servicesIndex *servicesindex.ServicesIndex,
) error {
	sourcesByPort := make(map[string][]string)
	addSource := func(port, source string) {
		// a service can be required by more than one brick
		if !slices.Contains(sourcesByPort[port], source) {
			sourcesByPort[port] = append(sourcesByPort[port], source)
		}
	}

	for _, appBrick := range app.Bricks {
		indexBrick, found := index.FindBrickByID(appBrick.ID)
		if !found {
			continue
		}

		for _, p := range indexBrick.GetPorts() {
			addSource(p, appBrick.ID)
		}

		matchingServices, err := indexBrick.GetMatchingService(bricksindex.BrickInstance{
			Model: cmp.Or(appBrick.Model, indexBrick.ModelName),
		})
		if err != nil {
			return fmt.Errorf("failed to get required services for brick %s: %w", appBrick.ID, err)
		}

		for _, id := range matchingServices {
			service, found := servicesIndex.FindServiceByID(id)
			if !found {
				slog.Debug("service required by brick not found or not available for current board", slog.String("service_id", id), slog.String("brick_id", appBrick.ID))
				continue
			}
			for _, p := range service.GetPorts() {
				addSource(p, fmt.Sprintf("%s (service %s)", appBrick.ID, service.Name))
			}
		}
	}

	for _, p := range app.Ports {
		addSource(strconv.Itoa(p), appPortsSource)
	}

	var allErrors error
	for _, p := range slices.Sorted(maps.Keys(sourcesByPort)) {
		if len(sourcesByPort[p]) > 1 {
			allErrors = errors.Join(allErrors, fmt.Errorf("port %s is declared by multiple sources: %s", p, strings.Join(sourcesByPort[p], ", ")))
		}
	}

	return allErrors
}
```

lucarin91 (changes requested):

lucarin91 (approved):

PR #593, opened 2026-08-20
feat: download HF model by link

### LINKED ISSUE #565: Support for downloading custom llama.cpp (GGUF) models from HF
Given a link of a model in HF, it will be parsed and sent to the model_downloader to download, updating the model_list.yaml.

giulio93:
### Motivation

A user can install only the models that `models-list.yaml` declares. The runner can serve other GGUF models from Hugging Face. This change accepts a link to one of them.

### Change description

`POST /v1/models` downloads a model from a Hugging Face link. The body gives `model_url`, and `mmproj_url` for a vision model. The body gives no id: the downloader makes the id from the file that it writes. The `done` event reports the model as the listing describes it.

`PUT /v1/models/{id}` still installs a model that `models-list.yaml` declares. It now answers from the declaration, so no container runs before the download. It answers 404 for an id that no entry declares. It no longer answers 409 for an installed model: the request is idempotent, and the stream ends with `done`.

A model id travels base64url encoded on all routes. An undeclared id holds the repository path, which a path segment cannot carry. Each response also gives the plain id as `id_decoded`, for display. Disk and `app.yaml` keep the plain id.

`status` gains `downloading`, from the handler's `.download` marker. `metadata` gains `source-model-url`, from the download record.

`models-list.yaml` declares what a model is. The listing container reports its state. Each question about state starts that container. `checkBricks` and the bricks page paid one run for each brick. A caller with more than one question now holds a `Lookup`, which runs the listing one time. Only a `Lookup` starts the container, so the cost is visible at the call site.

`GET /v1/models` answered a partial list after a listing failure, with every model as `not-installed`. It now answers 500.

### Additional Notes

Breaking: the plain id fails on the wire, `status` has a new value, and `GET /v1/models`
fails after a listing failure.

### Reviewer checklist


internal/api/docs/openapi.yaml
```diff
@@ -1125,6 +1125,68 @@ paths:
       summary: Download and install a custom Edge Impulse AI model
       tags:
       - AIModels
+  /v1/models/hf:
```
lucarin91: Not sure if we should add a new endpoint for that; maybe we should use the same one we already use for the curated model

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -174,15 +192,67 @@ func (m *ModelsIndex) GetModelsByBrick(brickID string) []AIModelLite {
...
+func (m *ModelsIndex) GetModels(ctx context.Context) []AIModel {
+	models, err := m.listModels(ctx)
+	if err != nil {
+		slog.Warn("cannot get models info, falling back to declared models", "err", err)
+	}
+	return models
+}
+
+func (m *ModelsIndex) GetModelByID(ctx context.Context, id string) (*AIModel, error) {
+	return m.NewLookup().ByID(ctx, id)
```
mirkoCrobu, line 320: `GetModelByID `is used in a loop inside `getAppEnvironmentVariables`, so app start makes one container request per brick.

Something like this inside getAppEnvironmentVariables would do:


models := modelsIndex.NewLookup()      // just once for the whole func
for _, brick := range app.Descriptor.Bricks {
    ...
    if brick.Model != "" {             // no model, no question to the container
        if m, err := models.ByID(ctx, brick.Model); err != nil {
    ...
}

mirkoCrobu:
About the listing cost: 

The **full model list is read** by `GET /v1/models`, `GET /v1/models/{id}`, `GET /v1/bricks/{id}`, `GET /v1/apps/{id}/bricks[/{id}]`, `brick create/update validation`, `app start`, and the` CLI model list`.

 **It changes on only three**: `PUT /v1/models/{id}`, `DELETE /v1/models/{id}`, and `PUT /v1/models/ei/projects/{id}`.

So why not keep the full list cached at process level, built on first use, and invalidate it only on those three? Every read path would then be free, and Lookup would no longer be needed: it's the same idea, just scoped to a single request.

I'd build it lazily rather than at boot, so the daemon doesn't have to wait on Docker at startup for something a user may never ask for.

Maybe could be done in a different PR

internal/orchestrator/modelsindex/models_index.go
giulio93: Integrated

giulio93:
> About the listing cost:
> 
> The **full model list is read** by `GET /v1/models`, `GET /v1/models/{id}`, `GET /v1/bricks/{id}`, `GET /v1/apps/{id}/bricks[/{id}]`, `brick create/update validation`, `app start`, and the` CLI model list`.
> [...]

 The invalidation set is smaller than the set of things that change the list, though. downloading comes from the on-disk .download marker, so the list changes during a download and after a crash, not only when a PUT returns. arduino-app-cli model delete runs the orchestrator in its own process against the same models dir, so the daemon never sees it. Same for a model directory removed by hand on the board. And loadCustomModels reads the custom-models dir on every call today, so an EI model appearing or disappearing is currently picked up for free and a cache would freeze it. The daemon is long-lived, so anything stale stays stale until a restart.

 So the cache needs an invalidation that does not depend on who did the writing. The exact one is a watch on its inputs, the custom-models dir and the llamacpp tree, which is the same shape as the network watch. A TTL would also do if we accept a window.

With that in place Lookup does collapse into the cache, you're right. Until then it is the same idea scoped to one request, which is the part that needs no invalidation story to be correct.

internal/api/api.go
```diff
@@ -64,6 +64,7 @@ func NewHTTPRouter(
 	mux.Handle("GET /v1/models", handlers.HandleModelsList(modelsIndex))
 	mux.Handle("GET /v1/models/{modelID}", handlers.HandlerModelByID(modelsIndex))
 	mux.Handle("PUT /v1/models/ei/projects/{projectID}", handlers.HandleInstallEIModel(cfg, bricksIndex, modelsIndex, dockerClient, platform))
+	// modelID is either a declared model id or a Hugging Face source to fetch one from.
```
lucarin91: I think we need to have a POST endpoint to download models from URLs. It will be strange to use a PUT on an external resource.

```
mux.Handle("POST /v1/models", 
```

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -174,15 +192,67 @@ func (m *ModelsIndex) GetModelsByBrick(brickID string) []AIModelLite {
...
+}
+
+// GetModelsByBrick returns the models that are associated with the given brick name.
+func (m *ModelsIndex) GetModelsByBrick(ctx context.Context, brickID string) []AIModelLite {
+	matches, err := m.NewLookup().ByBrick(ctx, brickID)
+	if err != nil {
+		slog.Warn("cannot get models info, brick compatibility list may be incomplete", "brick", brickID, "err", err)
+	}
```
lucarin91, line 328: Those old methods should be removed; otherwise, we could use them by mistake

internal/orchestrator/modelsindex/models_index.go
```diff
@@ -174,15 +192,67 @@ func (m *ModelsIndex) GetModelsByBrick(brickID string) []AIModelLite {
...
 	return matches
 }
 
-func (m *ModelsIndex) IsModelSupportedByBrick(modelID, brickID string) bool {
-	models := m.GetModelsByBrick(brickID)
-	return slices.ContainsFunc(models, func(model AIModelLite) bool {
-		return model.ID == modelID
-	})
+func (m *ModelsIndex) IsModelSupportedByBrick(ctx context.Context, modelID, brickID string) (bool, error) {
+	return m.NewLookup().SupportedByBrick(ctx, modelID, brickID)
```
lucarin91: also this

cmd/gendoc/docs.go
```diff
@@ -875,7 +875,7 @@ Contains a JSON object with the details of an error.
 			Method:      http.MethodDelete,
 			Path:        "/v1/models/{id}",
 			Request: (*struct {
-				ID    string `path:"id" description:"AI model identifier"`
+				ID    string `path:"id" description:"AI model identifier. An id no internal model list entry declares carries the repository path the model was downloaded from, so its slashes must be percent-encoded." example:"llamacpp:unsloth%2FSmolLM2-135M-Instruct-GGUF%2FSmolLM2-135M-Instruct-Q4_K_M"`
```
lucarin91: We are provising the ids after the hf model installation, so we don't need to be specific about the format

```suggestion
				ID    string `path:"id" description:"AI model identifier."`
```

cmd/gendoc/docs.go
```diff
@@ -897,7 +897,7 @@ Contains a JSON object with the details of an error.
 			Method:      http.MethodGet,
 			Path:        "/v1/models/{id}",
 			Request: (*struct {
-				ID string `path:"id" description:"AI model identifier."`
+				ID string `path:"id" description:"AI model identifier. An id no internal model list entry declares carries the repository path the model was downloaded from, so its slashes must be percent-encoded." example:"llamacpp:unsloth%2FSmolLM2-135M-Instruct-GGUF%2FSmolLM2-135M-Instruct-Q4_K_M"`
```
lucarin91:
```suggestion
				ID string `path:"id" description:"AI model identifier."`
```

cmd/gendoc/docs.go
```diff
@@ -938,6 +938,67 @@ Contains a JSON object with the details of an error.
...
+			// Documented as {id} like the GET and DELETE on the same path: OpenAPI treats
+			// two templated paths of the same hierarchy with different names as identical,
+			// so {modelID} here would collide with them.
+			Path: "/v1/models/{id}",
+			Parameters: (*struct {
+				ModelID string `path:"id" description:"A model id the internal model list declares, or a Hugging Face source to fetch one from: a file URL, or the compact \"[type:]repo:quantization[:mmproj_quantization]\" key. Percent-encode the slashes." example:"llamacpp:unsloth%2FSmolLM2-135M-Instruct-GGUF:Q4_K_M"`
+			})(nil),
+			Request: (*struct {
+				MmprojURL string `json:"model_mmproj_url" description:"Multimodal projection file for a vision model, when the source is a file URL. A compact key names it inline instead, and a declared model needs no body at all." example:"https://huggingface.co/unsloth/SmolLM2-135M-Instruct-GGUF/blob/main/mmproj-F16.gguf"`
+			})(nil),
```
lucarin91, line 964: As I wrote below, this should be split into a PUT with a path parameter and a POST with a body parameter

internal/orchestrator/modelsindex/models_index.go
lucarin91, line 77: Now that a model is either an internal reference or an external one, we should be clear about its format, and maybe we could create a helper type like the app id one https://github.com/arduino/arduino-app-cli/blob/main/internal/orchestrator/appid/id.go#L21-L26

lucarin91 (dismissed):

internal/api/docs/openapi.yaml
```diff
@@ -1071,6 +1071,84 @@ paths:
...
+          source to fetch one from: a file URL, or the compact "[type:]repo:quantization[:mmproj_quantization]"
+          key. Percent-encode the slashes.'
+        in: path
+        name: id
+        required: true
+        schema:
+          description: 'A model id the internal model list declares, or a Hugging
+            Face source to fetch one from: a file URL, or the compact "[type:]repo:quantization[:mmproj_quantization]"
+            key. Percent-encode the slashes.'
+          example: llamacpp:unsloth%2FSmolLM2-135M-Instruct-GGUF:Q4_K_M
```
lucarin91: This is ambiguous; we need at least to prefix it with `hf`, but I would also say that we can remove the support of compat ID if not needed.

lucarin91:
Some improvements:
- [x] We should return the original download URL

cmd/gendoc/docs.go
```diff
@@ -875,7 +892,7 @@ Contains a JSON object with the details of an error.
 			Method:      http.MethodDelete,
 			Path:        "/v1/models/{id}",
 			Request: (*struct {
-				ID    string `path:"id" description:"AI model identifier"`
+				ID    string `path:"id" description:"AI model identifier. Send back either form a response reports: \"id\", which is base64url encoded and unpadded, or \"id_decoded\", which is plain - both name the same model. An id holding slashes has to be the encoded form, since percent-encoding it is refused." example:"bGxhbWFjcHA6dW5zbG90aC9TbW9sTE0yLTEzNU0tSW5zdHJ1Y3QtR0dVRi9TbW9sTE0yLTEzNU0tSW5zdHJ1Y3QtUTRfS19N"`
```
lucarin91:
```suggestion
				ID    string `path:"id" description:"AI model identifier.`
```

cmd/gendoc/docs.go
```diff
@@ -897,7 +914,7 @@ Contains a JSON object with the details of an error.
 			Method:      http.MethodGet,
 			Path:        "/v1/models/{id}",
 			Request: (*struct {
-				ID string `path:"id" description:"AI model identifier."`
+				ID string `path:"id" description:"AI model identifier. Send back either form a response reports: \"id\", which is base64url encoded and unpadded, or \"id_decoded\", which is plain - both name the same model. An id holding slashes has to be the encoded form, since percent-encoding it is refused." example:"bGxhbWFjcHA6dW5zbG90aC9TbW9sTE0yLTEzNU0tSW5zdHJ1Y3QtR0dVRi9TbW9sTE0yLTEzNU0tSW5zdHJ1Y3QtUTRfS19N"`
```
lucarin91:
```suggestion
				ID string `path:"id" description:"AI model identifier.`
```

cmd/gendoc/docs.go
```diff
@@ -938,6 +955,114 @@ Contains a JSON object with the details of an error.
 				{StatusCode: http.StatusInsufficientStorage, Reference: "#/components/responses/InsufficientStorage"},
 			},
 		},
+		{
+			OperationId: "installModel",
+			Method:      http.MethodPut,
+			// Documented as {id} like the GET and DELETE on the same path: OpenAPI treats
+			// two templated paths of the same hierarchy with different names as identical,
+			// so {modelID} here would collide with them.
```
lucarin91:
```suggestion
```

cmd/gendoc/docs.go
```diff
@@ -938,6 +955,114 @@ Contains a JSON object with the details of an error.
...
+			Path: "/v1/models/{id}",
+			Parameters: (*struct {
+				ModelID string `path:"id" description:"The id of a model in the internal model list, as either form a response reports: the base64url encoded \"id\" or the plain \"id_decoded\"." example:"bGxhbWFjcHA6Z2VtbWEtMy0xYi1pdC1RNF8w"`
```
lucarin91:
```suggestion
				ModelID string `path:"id" description:"The id of a model in the internal model list.`
```

cmd/gendoc/docs.go
```diff
@@ -938,6 +955,114 @@ Contains a JSON object with the details of an error.
...
+every later failure is an error event. A models directory with no free space has the code
+'insufficient_storage'.
+'event: error'
+'data: {"code":"INTERNAL_SERVER_ERROR","message":"An error occurred during operation"}'
+'data: {"code":"insufficient_storage","message":"insufficient disk space to install model"}'
+`,
+			},
+			Description: `Install an AI model from the internal model list. The progress is a stream of Server-Sent Events.
+
+The path takes a model id, not a download source. An id that no entry declares gives a 404: to download a Hugging Face model, use POST /v1/models. No container starts before the download, because the check reads the model list only.
```
lucarin91:
```suggestion
The path takes a model ID from the internal model list. An invalid model ID returns 404.
```

cmd/gendoc/docs.go
```diff
@@ -938,6 +955,114 @@ Contains a JSON object with the details of an error.
...
+**Event 'error'**:
+Contains a JSON object with the details of an error. The 200 status is sent with the stream, so
+every later failure is an error event. A models directory with no free space has the code
+'insufficient_storage'.
+'event: error'
+'data: {"code":"INTERNAL_SERVER_ERROR","message":"An error occurred during operation"}'
+'data: {"code":"insufficient_storage","message":"insufficient disk space to install model"}'
+`,
+			},
+			Description: `Download an AI model from Hugging Face. The progress is a stream of Server-Sent Events.
```
lucarin91, line 1047:
```suggestion
			Description: `Download an LLamaCPP AI model from a Hugging Face link. The progress is a stream of Server-Sent Events.
```

cmd/gendoc/docs.go
```diff
@@ -938,6 +955,114 @@ Contains a JSON object with the details of an error.
...
+
+"model_url" is the URL of the model file on Hugging Face. It selects one file at one commit. For a vision model, "model_mmproj_url" is the URL of the projection file. Only llama.cpp models are supported: the file must be a GGUF file, and it goes in the llamacpp models directory.
+
+The request has no model id. The downloader makes the id from the file that it writes, and reports it in the "done" event. If the internal model list declares that file, the answer is the declared model.
```
lucarin91, line 1051:
```suggestion
The downloader makes the id from the file that it writes, and reports it in the "done" event. If the internal model list declares that file, the answer is the declared model.
```

cmd/gendoc/docs.go
```diff
@@ -938,6 +955,114 @@ Contains a JSON object with the details of an error.
...
+
+Hugging Face reads the URL at the download only, so a bad URL is an error event. The request is idempotent: the handler does not transfer a file that is on disk again.
+
+The models-downloader image must contain arduino/app-bricks-py#415 and arduino/app-bricks-py#416, with the llamacpp-runner images from the same release. An older image installs the model, but reports no id, and the stream sends an error event.`,
```
lucarin91, line 1055:
```suggestion
`,
```

cmd/gendoc/docs.go
```diff
@@ -938,6 +955,114 @@ Contains a JSON object with the details of an error.
...
+			Summary: "Download a llama.cpp model from Hugging Face",
```
lucarin91, line 1056:
```suggestion
			Summary: "Download a LLamaCPP model from Hugging Face",
```

internal/api/handlers/ai_models.go
```diff
@@ -46,11 +46,38 @@ func HandleModelsList(modelsIndex *modelsindex.ModelsIndex) http.HandlerFunc {
...
+// or by the base64url encoding of it, so a client holding either form is right and this
+// handler needs to know nothing about which. Older clients kept the plain id; a response
+// now reports the encoded one as "id".
+//
+// A percent-encoded id is the one form refused. The router turns %2F into a real slash
+// before the handler runs, so accepting it would make the request depend on nothing in
+// the path normalizing it, which this API cannot promise.
+// errEmptyModelID is its own error because the routes answer it with different statuses,
+// and did so before this parsing was shared: delete reports 412, as every other route
+// that takes an id in a path does, and the rest report 400.
```
lucarin91, line 62:
```suggestion
```

internal/api/handlers/ai_models.go
```diff
@@ -46,11 +46,38 @@ func HandleModelsList(modelsIndex *modelsindex.ModelsIndex) http.HandlerFunc {
...
+var errEmptyModelID = errors.New("id must be set")
+
+func modelIDFromPath(r *http.Request) (string, error) {
+	segment := strings.TrimSpace(r.PathValue("modelID"))
+	if segment == "" {
+		return "", errEmptyModelID
+	}
+	if strings.Contains(segment, "/") {
+		return "", errors.New("a model id holding slashes travels base64url encoded, not percent-encoded")
+	}
```
lucarin91, line 72: I don't think we can check much on the base64 ids
```suggestion
```

internal/api/handlers/ai_models.go
```diff
@@ -169,26 +200,81 @@ func (r InstallEIModelRequest) Validate() error {
...
+// HandleInstallModel installs a model from the internal model list. HandleDownloadModel
+// downloads a model that no entry declares.
 func HandleInstallModel(dockerClient command.Cli, modelsIndex *modelsindex.ModelsIndex, plat platform.Platform) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		id := strings.TrimSpace(r.PathValue("modelID"))
-		if id == "" {
-			render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "model ID must be set"})
+		id, err := modelIDFromPath(r)
+		if err != nil {
+			render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: err.Error()})
```
lucarin91, line 220: You can now remove `modelIDFromPath` and revert this code


```suggestion
id := strings.TrimSpace(r.PathValue("modelID"))
		if id == "" {
			render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "model ID must be set"})
```

internal/api/handlers/ai_models.go
```diff
@@ -169,26 +200,81 @@ func (r InstallEIModelRequest) Validate() error {
...
 			return
 		}
 
-		model, err := modelsIndex.GetModelByID(r.Context(), id)
+		// The declaration alone answers this, so no listing container runs.
+		declared, found := modelsIndex.DeclaredByID(id)
+		if !found {
+			details := fmt.Sprintf("no model with id %q is declared; download a Hugging Face model with POST /v1/models", id)
```
lucarin91, line 227:
```suggestion
			details := fmt.Sprintf("no model with id %q is declared", id)
```

internal/api/handlers/ai_models.go
```diff
@@ -169,26 +200,81 @@ func (r InstallEIModelRequest) Validate() error {
...
+
+		// The second result is false only when the declaration is nil.
+		installed, _ := installedModel(modelsIndex, declared, stream.downloaded)
+		sseStream.Send(render.SSEEvent{Type: "done", Data: orchestrator.NewAIModelItem(installed)})
+	}
+}
+
+type DownloadModelRequest struct {
+	ModelURL  string `json:"model_url" description:"URL of the GGUF model file on Hugging Face" example:"https://huggingface.co/unsloth/SmolLM2-135M-Instruct-GGUF/resolve/main/SmolLM2-135M-Instruct-Q4_K_M.gguf" required:"true"`
+	MmprojURL string `json:"model_mmproj_url" description:"URL of the GGUF multimodal projection file on Hugging Face, for a vision model" example:"https://huggingface.co/unsloth/SmolLM2-135M-Instruct-GGUF/resolve/main/mmproj-F16.gguf"`
```
lucarin91:
```suggestion
	MmprojURL string `json:"mmproj_url" description:"URL of the GGUF multimodal projection file on Hugging Face, for a vision model" example:"https://huggingface.co/unsloth/SmolLM2-135M-Instruct-GGUF/resolve/main/mmproj-F16.gguf"`
```

internal/api/handlers/ai_models.go
```diff
@@ -200,47 +286,93 @@ func HandleInstallModel(dockerClient command.Cli, modelsIndex *modelsindex.Model
...
-					Message: "insufficient disk space to install model",
-				})
-				return
-			}
+		installed, ok := installedModel(modelsIndex, nil, stream.downloaded)
+		if ok && installed.Source == nil {
+			// The links the caller asked for, so the model this event describes carries
+			// the same source the listing will report for it. The recorded timestamp is
+			// not on the stream, and inventing one here would disagree with the record.
+			installed.Source = &modelsindex.ModelSource{ModelURL: modelURL, MmprojURL: strings.TrimSpace(req.MmprojURL)}
```
lucarin91: If this information is not coming from the listing of the model, it doesn't make sense to add that here. Rather, ask Wire to store and/or expose that.
```suggestion
```

internal/api/handlers/ai_models.go
```diff
@@ -200,47 +286,93 @@ func HandleInstallModel(dockerClient command.Cli, modelsIndex *modelsindex.Model
...
+		}
+		if !ok {
+			slog.Error("download named no model", "model_url", modelURL)
```
lucarin91:
```suggestion
```

internal/orchestrator/bricks/bricks.go
```diff
@@ -379,10 +391,16 @@ func (s *Service) BrickCreate(
...
+		model, err := s.modelsIndex.ModelForBrick(ctx, *req.Model, req.ID)
+		if err != nil {
+			return fmt.Errorf("checking model %s: %w", *req.Model, err)
+		}
+		if model == nil {
 			return fmt.Errorf("model %s does not exsist", *req.Model)
 		}
-		brickInstance.Model = *req.Model
+		// The model's own id, not the string the request carried: a model id travels
+		// either plainly or base64url encoded, and app.yaml keeps only the plain form.
```
lucarin91:
```suggestion
```

internal/orchestrator/bricks/bricks.go
```diff
@@ -420,10 +439,16 @@ func (s *Service) BrickUpdate(
...
+		model, err := s.modelsIndex.ModelForBrick(ctx, *req.Model, req.ID)
+		if err != nil {
+			return fmt.Errorf("checking model %s: %w", *req.Model, err)
+		}
+		if model == nil {
 			return fmt.Errorf("model %s is not supported by brick %q", *req.Model, req.ID)
 		}
-		brickModel = *req.Model
+		// See AddBrick: the stored id is the model's, so an encoded request and a plain
+		// one leave the same app.yaml behind.
```
lucarin91:
```suggestion
```

internal/orchestrator/modelsindex/models_index.go
giulio93: removed https://github.com/arduino/arduino-app-cli/pull/593/commits

internal/orchestrator/modelsindex/models_index.go
giulio93: Agreed. [41b1a221](https://github.com/arduino/arduino-app-cli/pull/593/changes/41b1a22125835d766eaa310c7eb9d71caca2a1ed) settles the format half: a model id is base64url everywhere in the API, and DecodeID is the only way in, called both on the {id} path segment and on a brick request's model.

The type I'd do as a follow-up. It would be weaker than appid.ID though: IDFromBase64 checks the decoded path is in a known location, so holding an appid.ID already means the app exists. We can't check that for a model, because whether an id names a real model is only answered by the catalog or by the listing container. A modelid.ID could only check that the string is valid base64url, which DecodeID does today. It would still be worth having, to stop the encoded and plain forms being mixed up by accident.

internal/api/docs/openapi.yaml
giulio93: Resolved by the endpoint split. The ambiguity was one path parameter meaning two things; now PUT /v1/models/{id} takes a declared id only and 404s on anything else, and a Hugging Face source goes to POST /v1/models as a full model_url in the body.

internal/orchestrator/modelsindex/handlers_index.go
```diff
@@ -176,16 +176,110 @@ type handlerModelListOutput struct {
 	Models []handlerModelEntry `json:"models"`
 }
 
+type entryMetadata struct {
+	ModelID      string            `json:"model_id"`
+	Handler      string            `json:"handler"` // a handler id, e.g. "hf-handler"
+	DownloadedAt string            `json:"downloaded_at"`
```
mirkoCrobu: This is dead code: used just in test

mirkoCrobu (approved):

PR #594, opened 2026-08-21
docs(ei model specification): Use verbatim EI category names in compatibility table

per1234:
### Motivation

The determination of which bricks an Edge Impulse model may be used with is based on the name of the project category assigned by Edge Impulse.

The mapping between category name and Brick IDs is documented by a table in the "Edge Impulse Model Specification". This information is essential for the user to ensure their model can be used with a given Brick, and for troubleshooting the situation where a model is unexpectedly not available for selection from the Brick configuration in App Lab.

Previously, there was some inconsistency between the category names as written in the table and the actual names used by Edge Impulse. This made it impossible for the user to determine from the table alone whether their project category would result in a model compatible with a target Brick. They would instead only be able to determine this by trial and error, or examining the Arduino App CLI source code.

### Change description

The category names written in the compatibility table are hereby adjusted to exactly match the actual names.

### Reviewer checklist


mirkoCrobu (approved):
Nice improvement

PR #598, opened 2026-08-26
fix: handle sigterm signal for systemd

lucarin91:
### Motivation

Our graceful shutdown mechanism wasn't handling the SIGTERM signal used by systemd, and recently also by us. 

This is especially important in the update flow, when we signal ourselves to restart, and we will need to make sure all the inflight messages to the FE are actually sent.

### Change description

This PR also cover all the following changes to be sure the daemon server is correctly terminated.

1. Add a custom context in the handler so we can kill the ongoing SSE stream
1. Move the daemon self-kill in the update manager so we are sure restart events are always sent
1. Make sure all the channels, including SSE and update broadcast ones, are flushed before termination

So the termination flow is:

1. The daemon receives SIGINT 
2. We signal termination with a new context 
4. All ongoing requests are terminated, including the SSE ones
5. At most after 30 seconds, the whole server is stopped

If the signal is sent by themself for autotermination, we are sure that all the channels are flushed and so the SSE caller got the restart|done event.

This PR also covers all the following changes to be sure the daemon server is correctly terminated.

1. Handle SIGTERM, and not only SIGINT, in the root context of the CLI
1. Give the HTTP server a custom `BaseContext`, so we can kill the ongoing SSE streams
1. Move the daemon self-kill from the apt service to the update manager, so the restart and done events are always broadcast before the signal
1. Flush the events already broadcast to the SSE clients before their streams are closed

So the termination flow is:

1. The daemon receives SIGTERM, or SIGINT when it runs in a terminal
2. The root context is cancelled, and with it the context of every ongoing request
3. Each SSE stream delivers the events it already has, then it is closed with the `close` event
4. The server waits for the handlers, 30 seconds at most; with the streams cut, it takes milliseconds

If the signal is sent by ourselves for autotermination, every event is broadcast before the signal is raised, so the SSE caller always gets the `restarting` and `done` events.

### Reviewer checklist

### DIFF
--- a/cmd/arduino-app-cli/daemon/daemon.go
+++ b/cmd/arduino-app-cli/daemon/daemon.go
@@ -138,7 +138,7 @@ func httpHandler(ctx context.Context, cfg config.Configuration, daemonPort, vers
 
 	// Start the HTTP server
 	address := "127.0.0.1:" + daemonPort
-	// All the requests derive from srvCtx: cancelling it interrupts the long-lived
+	// All the requests derive from srvCtx: canceling it interrupts the long-lived
 	// SSE streams, that otherwise block the shutdown until the timeout.
 	srvCtx, cancelRequests := context.WithCancel(context.Background())
 	httpSrv := http.Server{
--- a/cmd/arduino-app-cli/daemon/daemon.go
+++ b/cmd/arduino-app-cli/daemon/daemon.go
@@ -138,7 +138,7 @@ func httpHandler(ctx context.Context, cfg config.Configuration, daemonPort, vers
 
 	// Start the HTTP server
 	address := "127.0.0.1:" + daemonPort
-	// All the requests derive from srvCtx: canceling it interrupts the long-lived
+	// All the requests derive from srvCtx: canceling it interrupts all requests, including the long-lived
 	// SSE streams, that otherwise block the shutdown until the timeout.
 	srvCtx, cancelRequests := context.WithCancel(context.Background())
 	httpSrv := http.Server{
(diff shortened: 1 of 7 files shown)


lucarin91:
~We should kill the SSE connection when we receive the signal; otherwise, a simple `curl http://localhost:8800/v1/system/resources` will block the server shutdown till the timeout ends~

done

cmd/arduino-app-cli/daemon/daemon.go
```diff
@@ -137,10 +138,14 @@ func httpHandler(ctx context.Context, cfg config.Configuration, daemonPort, vers
 
 	// Start the HTTP server
 	address := "127.0.0.1:" + daemonPort
+	// All the requests derive from srvCtx: canceling it interrupts the long-lived
+	// SSE streams, that otherwise block the shutdown until the timeout.
```
mirkoCrobu:
```suggestion
	// All the requests derive from srvCtx: canceling it interrupts all requests, including the long-lived
	// SSE streams, that otherwise block the shutdown until the timeout.
```

internal/update/update.go
mirkoCrobu, line 241: The PR looks good to me. The last place where we can actually guarantee `restarting `and `done `reach the client is probably broadcast itself: with a buffer of 100 and those two always being the last events. 
We could handle terminal events explicitly. What do you think?

lucarin91: I don't understand. Are you afraid that those two messages will not be in the channel in the first place? What is your proposal then?

martacarbone (approved):

PR #601, opened 2026-08-26
feat(task): publish a locally built package as an update on the board

### LINKED ISSUE #596: Add a task command to publish a locally built package as an update on the board
Testing the system update flow on a board currently takes a manual setup for the scenario where we want to test a build as available upgrade. 

**Manual steps:** 
 build the package, give it a version that outranks the released one, copy it over, turn the destination into an apt repository and register it as a source. 

The task command should automate that setup so that, once run, a `system update` on the board offers the locally built version as the available upgrade.

**What it does**

1. Builds the debian package from the current working tree.
2. Assigns it a version high enough to win over whatever the official repository offers, so apt picks the local build as the candidate.
3. Copies it to the board, into a location that survives a reboot.
4. Generates the package index next to it and registers the directory as an apt source, so apt treats it as a normal repository.

After that the board is ready: triggering an update through the daemon, the CLI, or the app installs the code under test.

**Notes:**

- The package must come from the working tree, not from a release: the update path also depends on files shipped by the package itself, so both ends of the upgrade have to carry the code under test.
- Worth a companion task to remove the source and the directory, to put the board back to following the official repository.

mirkoCrobu:
Closes #596

Adds `board:publish-update`: builds the working tree as a `.deb` with a version that
outranks the official one, copies it to the board, generates the apt index next to it
and registers it as an apt source, so a `system update` installs the code under test.

## Usage

```
go tool task board:publish-update          # publish
go tool task board:publish-update:clean    # remove it, back to the official repo
```

The board is reached over `adb` by default. Set `BOARD` for ssh, and `BOARD_PASSWORD`
(needs `sshpass`) to skip the password prompts:

```
BOARD=arduino@my-board.local BOARD_PASSWORD=... go tool task board:publish-update
```

The index is generated in Docker, not on the board, which has no `dpkg-dev`.
`board:install` is untouched.

Tested on an UNO Q: candidate becomes the local build and `apt list --upgradable`
reports it, so the daemon offers it.

### Reviewer checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,9 @@ go.work.sum
 
 build/
 
+# local task configuration, may hold the board password
+.env.local
+
 **/.cache
 
 /arduino-app-cli
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -260,6 +260,68 @@ tasks:
       - adb push ./build/arduino-app-cli_*_arm64.deb /tmp/
       - ./scripts/run-remote-sudo.sh "dpkg -i /tmp/arduino-app-cli*_arm64.deb && needrestart -r a"
 
+  # Publishes the working tree on the board as the update that `system update`
+  # will offer, by turning a locally built package into an apt repository the
+  # board follows. The package must come from the working tree and not from a
+  # release, because the update path depends on files shipped by the package
+  # itself: both ends of the upgrade have to carry the code under test.
+  #
+  # The board is reached over adb by default; set BOARD (e.g.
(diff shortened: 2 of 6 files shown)


lucarin91 (commented):
I am wondering if we can start the apt server locally and make the board connect to it instead of running that on the board. I am afraid that the current implementation will leave staff on the board that will broke normal upgrades. I think it will be more simpler to have a way to start a local repository maybe directly in the build dir and then have a way to share that with the board, via adb forward for instance.

PR #607, opened 2026-08-28
Report python runner version

robgee86:
### Motivation

`arduino-app-cli version` only reports the CLI and daemon versions. There is no way to see which Python runner image is in use without peeking at config files.

### Change description

- `GET /v1/version` now returns `python_runner_version` with the daemon's full Python runner image reference.
- `arduino-app-cli version` prints the runner image resolved from its own environment (`python_runner_version`) and the daemon's (`daemon_python_runner_version`).
- OpenAPI spec and e2e client updated accordingly.

### Additional Notes

N/A

### Reviewer checklist

### DIFF
--- a/cmd/arduino-app-cli/main.go
+++ b/cmd/arduino-app-cli/main.go
@@ -71,7 +71,7 @@ func run(configuration cfg.Configuration) error {
 		properties.NewPropertiesCmd(configuration),
 		config.NewConfigCmd(configuration),
 		system.NewSystemCmd(configuration),
-		version.NewVersionCmd(Version),
+		version.NewVersionCmd(Version, configuration.PythonImage),
 		monitor.NewMonitorCmd(),
 		model.NewModelCmd(configuration),
 	)
--- a/cmd/arduino-app-cli/version/version.go
+++ b/cmd/arduino-app-cli/version/version.go
@@ -26,9 +26,9 @@ const (
 	ProgramName     = "Arduino App CLI"
 )
 
-// NewVersionCmd creates the version command. bricksVersion is the Python runner
+// NewVersionCmd creates the version command. pythonRunnerVersion is the Python runner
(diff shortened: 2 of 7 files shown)


mirkoCrobu:
The runner image is a value derived from the environment, so conceptually it's configuration rather than version info, and `/v1/config` already exposes `python_runner`(populated with the build-time constant). Before adding a field to `/v1/version`, it would be worth clarifying the use case: if it's "which runner is running on this device", fixing the existing `python_runner` is enough. 
What's the scenario you had in mind?

image
(In the image, just notice that python_runner is missing in the textual command)

robgee86:
The `/v1/config` and `config get` paths both lead to a variable that is frozen at compile time, `cfg.RunnerVersion`.

My use-case is seeing which runner image the CLI/daemon resolved taking also in consideration overrides done by DOCKER_REGISTRY_BASE and DOCKER_PYTHON_BASE_IMAGE. This is useful in at least two scenarios:
- Most developers on app-bricks-py have the need to see which version is currently picked up in a simple, scriptable, way;
- Users who leverage this feature (both on purpose or if asked by support) may need to report their used version in the same simple way.

Agreed it's config, not version info, so we could move the feature to `/v1/config` and `config get` if we think it would be easy enough for users to report their current setup when support needs to triage.

mirkoCrobu:
I'd say we can expose the *resolved* runner image alongside the hardcoded `python_runner` in `/v1/config` and `config get`.
 In that way, we can expose the resolved runner image alongside the hardcoded 'python_runner' in '/v1/config' and 'config get'.
 This makes it easy to see what was baked at build time and what is actually resolved by DOCKER_REGISTRY_BASE and DOCKER_PYTHON_BASE_IMAGE. 
 
 Something like 'python_runner' (configured) + 'python_runner_resolved'? (Open to better suggestions).

PR #608, opened 2026-08-28
 include service ports into exposed-ports endpoint

mirkoCrobu:
### Motivation

Bricks can require a service through `requires_services`, and the service publishes its own ports.
The CLI never extracted them.

### Change description
For now we do the smaller thing: report the service ports in the current schema, unchanged, and
add the matching CLI command.

The accepted cost is that App Lab forwards each service port. That is one extra tunnel per service port,
- New `arduino-app-cli app ports <app_path>` command, with `--format json`.
- Unit tests on `AppPorts` and on the response mapping, plus an e2e case covering a required service.

### Reviewer checklist


mirkoCrobu:
### Conclusion after internal discussion

For now we do the smaller thing: **report the service ports in the current schema, unchanged, and
add the matching CLI command.** No new field, nothing for App Lab to migrate.

The accepted cost is that App Lab forwards each service port, since any value other than `webview`
falls into its `forwardNonUIPort` branch. That is one extra tunnel per service port, held for the
board session: waste, not breakage. We take it for the time being.

PR #609, opened 2026-08-28
Implement secret management

lucarin91:
### Motivation

The value of a variable a brick declares `secret` lives in `app.yaml`, inside the app folder.
Distributing an app release means also shipping that folder, so a secret cannot be kept there.

### Change description

A secret is now stored outside the app folder, in `<data-dir>/secrets/<hash-of-app-id>/`, one file per variable, `0600` in a `0700` directory.

- New `internal/orchestrator/secrets` package: `Get`/`Set`/`Delete`/`Move`/`Copy`, plus `Adopt`, which moves what the api wrote in `app.yaml` into the store and blanks the keys there.
- Adopt is called every time `StartApp` runs.
- A secret never reaches a template: the build-time environment only references it as`${NAME}`.
- The rendered compose file is the only file holding a resolved secret and is now `0600`.
- The store follows the app folder, which is what its id is made of: delete removes the secrets, rename moves them, clone copies them.

### DIFF
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -37,6 +37,7 @@ import (
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/modelsindex"
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/peripherals"
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/pipewire"
+	"github.com/arduino/arduino-app-cli/internal/orchestrator/secrets"
 	"github.com/arduino/arduino-app-cli/internal/orchestrator/servicesindex"
 	"github.com/arduino/arduino-app-cli/internal/platform"
 )
@@ -107,6 +108,21 @@ func StartApp(
 	}
 
 	bricksIndex = bricksIndex.WithAppBricks(appToStart.LocalBricks)
+
+	// The secrets of the app, by the id its folder has on this board. What the api
+	// wrote in app.yaml is moved into the store here, and app.yaml is blanked.
+	appToStartID, err := appid.NewAppProvider(cfg, platform).IDFromPath(appToStart.FullPath)
+	if err != nil {
+		return fmt.Errorf("cannot identify the app to start: %w", err)
+	}
+	secretsStore := secrets.NewStore(cfg)
+	if err := secrets.Adopt(secretsStore, appToStartID, &appToStart, bricksIndex); err != nil {
+		return fmt.Errorf("cannot store the secrets of the app: %w", err)
+	}
+	storedSecrets, err := secretsStore.Get(appToStartID)
+	if err != nil {
+		return fmt.Errorf("cannot read the secrets of the app: %w", err)
+	}
 
 	if err := checkBricks(ctx, appToStart.Descriptor.Bricks, bricksIndex, modelsIndex); err != nil {
 		return err
@@ -219,7 +235,7 @@ func StartApp(
 		// What the template references, answered on this board: for a release the app
 		// half will come from the bundle instead of being resolved again here.
 		env := hostEnvironment(ctx, appToStart.FullPath, cfg).Merge(appEnv)
-		if err := provisioner.Render(ctx, &appToStart, env, appSecrets(appToStart, bricksIndex)); err != nil {
+		if err := provisioner.Render(ctx, &appToStart, env, appSecrets(appToStart, bricksIndex, storedSecrets)); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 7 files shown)


internal/orchestrator/orchestrator.go
```diff
@@ -108,6 +109,21 @@ func StartApp(
...
 	bricksIndex = bricksIndex.WithAppBricks(appToStart.LocalBricks)
 
+	// The secrets of the app, by the id its folder has on this board. What the api
+	// wrote in app.yaml is moved into the store here, and app.yaml is blanked.
+	appToStartID, err := appid.NewAppProvider(cfg, platform).IDFromPath(appToStart.FullPath)
+	if err != nil {
+		return fmt.Errorf("cannot identify the app to start: %w", err)
+	}
+	secretsStore := secrets.NewStore(cfg)
+	if err := secrets.Adopt(secretsStore, appToStartID, &appToStart, bricksIndex); err != nil {
```
mirkoCrobu, line 119: secrets.Adopt should be done after checkPortCollision? Maybe it is too early here

Xayton: I agree. An app that does not start (for example because of checkBricks or checkPortCollisions fail), will have its app.yaml modified anyway.

More in general, I fear that modifying app.yaml for all started apps is a change to big to put here.

Xayton: I move the Adopt code later (with this commit [c14d876](https://github.com/arduino/arduino-app-cli/pull/609/commits/c14d876db4d77703f38643e45e69d30e8ed82b18)).

Xayton (commented):
I would consider using the Adopt only when resolving an app (exporting an App Release). I would not change the current behavior when running apps.

Xayton:
A few other things to fix before eventually merging:

1. We need a way to clear a secret from the store. If you remove a brick secret from the app.yaml variables, the value will keep being taken from the store and injected.
2. If the secrets file has the wrong permissions (not 0600) we should fail with an error, not just Warn.
3. Old secrets (that have been removed from the app or the bricks) accumulate in the store.

internal/orchestrator/secrets/secrets.go
```diff
@@ -0,0 +1,208 @@
...
+
+type Store struct {
+	dir *paths.Path
+}
+
+func NewStore(cfg config.Configuration) *Store {
+	return &Store{dir: cfg.DataDir().Join("secrets")}
+}
+
+// Get is the values of an app. An app with no secrets stored is not an error.
```
cmaglie:
```suggestion
// Get returns a map with all the secrets of an app. An app with no secrets stored returns an empty map.
```

internal/orchestrator/secrets/secrets.go
```diff
@@ -0,0 +1,208 @@
...
+			return io.ReadAll(f)
+		}()
+		if err != nil {
+			return nil, err
+		}
+		if value == nil {
+			continue
+		}
+		values[file.Base()] = string(value)
+	}
```
cmaglie: Why a different file for each secret? Isn't it simple to write all secrets in a single file?

internal/orchestrator/secrets/secrets.go
```diff
@@ -0,0 +1,208 @@
...
+	return values, nil
+}
+
+// Set writes the updates of an app, one file each. A name set to an empty value is
+// removed, and an app left with nothing has no directory.
```
cmaglie:
```suggestion
// Set updates the secrets of an app, one file each. A name set to an empty value is
// removed, and an app left with nothing has no directory.
```

internal/orchestrator/secrets/secrets.go
```diff
@@ -0,0 +1,208 @@
...
+			return err
+		}
+	}
+
+	// The remove fails while the app has a secret left, which is the check we want.
+	_ = dir.Remove()
+	return nil
+}
+
+func (s *Store) Delete(id appid.ID) error {
```
cmaglie:
```suggestion
// Delete removes all secrets of an app.
func (s *Store) Delete(id appid.ID) error {
```

internal/orchestrator/secrets/secrets.go
```diff
@@ -0,0 +1,208 @@
...
+	return s.appDir(id).RemoveAll()
+}
+
+// Copy repeats the values of an app under another id, as a clone does.
```
cmaglie:
```suggestion
// Copy the secrets from one app to another.
```

internal/orchestrator/secrets/secrets.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return err
+	}
+	if len(values) == 0 {
+		return nil
+	}
+	return s.Set(to, values)
+}
+
+// Move carries the values of an app to another id, as a rename or an upgrade does.
```
cmaglie:
```suggestion
// Move the secrets from one app to another.
```

PR #612, opened 2026-08-31
feat(cli): add 'app ps' command to show live app status

### LINKED ISSUE #586: Refactor app list to make more usable (app ps)
`app list` is both the place that show examples apps and the status of them, e.g. `start`, `stop` ecc. With the new inclusing of `core-and-fundamenta` examples and the `bricks` examples, this command is very full of stuff, and it is almast unusable. 

I think we should start thinking to add a `app ps` command to show only active apps and add some filter on the app list to make that more usable.

kirbx01:
### Motivation

`app list` currently mixes two different concerns: browsing the app catalog (examples and user apps) and showing live status (running, stopped, failed, etc). As the catalog grew with `core-and-foundational` and `bricks` examples, the command became cluttered and hard to use, as reported in #586.

Needs a dedicated way to see only the apps that are actually live on the board as outlined in the issue for enhancement, without the catalog noise, similar to `docker ps`.

### Change description

I added a new `app ps` command that shows only apps that have been initialized at least once.

By default it shows apps with status `starting`, `running`, `stopping`, or `failed`. I added an `--all` / `-a` flag that also includes `stopped` apps. Apps that have never been started (`uninitialized`) are never shown, in any combination of flags.

The table output is `ID`, `NAME`, `STATUS`.

Implementation notes:
- I reused the existing `orchestrator.ListApps` with `ShowExamples: true`, `ShowApps: true`, `IncludeNonStandardLocationApps: true`, then filtered the results client-side by status. I didn't change the orchestrator layer, since `ListAppRequest.StatusFilter` only supports a single exact status match and `app ps` needs to match a set of statuses.
- I registered the command in `app.go`.
- I added `ps_test.go` covering the default filter and the `--all` filter, including the case where `uninitialized` apps must be excluded in both.

This PR does not touch `app list`. That refactor (splitting status out of `app list` into a pure catalog view) is the second half of the split proposed in #586, and will follow in a separate PR.

### Demo

Local smoke test against a Linux-only (no-sketch) app, showing the full lifecycle: nothing shown before start, `running` after start, hidden by default after stop, and reappearing under `--all`.

Screenshot From 2026-09-01 03-42-49

### Additional Notes

Refs #586

This adds a new user-facing command; a follow-up doc update in arduino/docs-content may be needed to document `app ps` in the CLI reference.

### Reviewer checklist

### DIFF
--- a/cmd/arduino-app-cli/app/ps.go
+++ b/cmd/arduino-app-cli/app/ps.go
@@ -1,0 +1,105 @@
+// This file is part of arduino-app-cli.
+//
+// SPDX-FileCopyrightText: Arduino s.r.l. and/or its affiliated companies
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+package app
+
+import (
+	"context"
+
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/spf13/cobra"
+
+	"github.com/arduino/arduino-app-cli/cmd/arduino-app-cli/internal/cmdutil"
+	"github.com/arduino/arduino-app-cli/cmd/arduino-app-cli/internal/servicelocator"
+	"github.com/arduino/arduino-app-cli/cmd/feedback"
+	"github.com/arduino/arduino-app-cli/internal/orchestrator"
+	"github.com/arduino/arduino-app-cli/internal/orchestrator/config"
+	"github.com/arduino/arduino-app-cli/internal/tablestyle"
+)
+
+// psVisibleStatuses are the statuses shown by default, i.e. any app that has
+// been initialized at least once and is not currently stopped.
+var psVisibleStatuses = map[orchestrator.Status]bool{
+	orchestrator.StatusStarting: true,
+	orchestrator.StatusRunning:  true,
+	orchestrator.StatusStopping: true,
+	orchestrator.StatusFailed:   true,
+}
+
+func newPsCmd(cfg config.Configuration) *cobra.Command {
+	var showAll bool
+
+	cmd := &cobra.Command{
+		Use:   "ps",
+		Short: "List the status of the Arduino apps on the board",
+		Long: "List the Arduino apps that have been initialized at least once, and their live status.\n" +
+			"By default, only starting, running, stopping and failed apps are shown. Use --all to also " +
+			"include stopped apps. Apps that have never been started are never shown; use 'app list' to " +
+			"browse the full catalog instead.",
+		Run: func(cmd *cobra.Command, args []string) {
+			psHandler(cmd.Context(), cfg, showAll)
+		},
+	}
+
+	cmd.Flags().BoolVarP(&showAll, "all", "a", false, "Also show stopped apps")
+	return cmd
+}
+
+func psHandler(ctx context.Context, cfg config.Configuration, showAll bool) {
+	res, err := orchestrator.ListApps(ctx,
+		servicelocator.GetDockerClient(),
+		orchestrator.ListAppRequest{
+			ShowExamples:                   true,
+			ShowApps:                       true,
+			IncludeNonStandardLocationApps: true,
+		},
+		servicelocator.GetAppIDProvider(),
+		servicelocator.GetBricksIndex(),
+		cfg,
+		servicelocator.GetPlatform(),
+	)
+	if err != nil {
+		feedback.Fatal(err.Error(), feedback.ErrGeneric)
+	}
+
+	feedback.PrintResult(appPsResult{
+		Apps: filterAppsByStatus(res.Apps, showAll),
+	})
+}
+
+// filterAppsByStatus keeps only apps that have been initialized at least once
+// (i.e. excludes StatusUninitialized), showing stopped apps only if showAll is set.
+func filterAppsByStatus(apps []orchestrator.AppInfo, showAll bool) []orchestrator.AppInfo {
+	res := make([]orchestrator.AppInfo, 0, len(apps))
+	for _, a := range apps {
+		if psVisibleStatuses[a.Status] || (showAll && a.Status == orchestrator.StatusStopped) {
+			res = append(res, a)
+		}
+	}
+	return res
+}
+
+type appPsResult struct {
+	Apps []orchestrator.AppInfo `json:"apps"`
+}
+
+func (r appPsResult) String() string {
+	t := table.NewWriter()
+	t.SetStyle(tablestyle.CustomCleanStyle)
+	t.AppendHeader(table.Row{"ID", "NAME", "STATUS"})
+
+	for _, app := range r.Apps {
+		t.AppendRow(table.Row{
+			cmdutil.IDToAlias(app.ID),
+			app.Name,
+			app.Status,
+		})
+	}
+	return t.Render()
+}
+
+func (r appPsResult) Data() any {
+	return r
+}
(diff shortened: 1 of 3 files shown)


kirbx01:
Hi @lucarin91 

I opened a PR implementing the first half of the split you outlined above: [PR #612 ](https://github.com/arduino/arduino-app-cli/pull/612) adds the `app ps` command.

It follows the layout you specified:
- `app ps` shows `starting` / `running` / `stopping` / `failed` apps
- `app ps --all` / `-a` also includes `stopped` apps
- `uninitialized` apps are excluded in all cases

The `app list` refactor (pure catalog view, `--examples` / `--all` filters, dropping the status/icon columns is the second half I'll follow up with a separate PR for that later today, as you suggested keeping the two efforts split.

Let me know if the `ps` output/flags look right, or if you'd want anything adjusted before the catalog PR follows.

cmd/arduino-app-cli/app/ps.go
```diff
@@ -0,0 +1,105 @@
...
+			psHandler(cmd.Context(), cfg, showAll)
+		},
+	}
+
+	cmd.Flags().BoolVarP(&showAll, "all", "a", false, "Also show stopped apps")
+	return cmd
+}
+
+func psHandler(ctx context.Context, cfg config.Configuration, showAll bool) {
+	res, err := orchestrator.ListApps(ctx,
```
lucarin91, line 51: I would recommend not using `ListApps` here, which is more useful for an app catalog. The reason is twofold:
1. We skip a useless filesystem scan; most applications on a board are uninitialized
2. We make sure we get all running/runned application also the one in unknown locations, with missing compose files, or even broken ones

I would suggest instead basing the command on a new function called `ListActiveApps` that internally calls `getAppsStatus` to get all the applications with a non-uninitialized status from Docker and convert them to an `AppInfo` array that can be used here.

lucarin91 (changes requested):
Thanks for the contribution, in the following comment, I added a change I would do to make the command more robust.

cmd/arduino-app-cli/app/ps.go
kirbx01: Okay so I worked on your suggestions and I changed `app ps` to use a new `ListActiveApps` function, which gets the app status directly from Docker via `getAppsStatus` and converts the results to `AppInfo`.

Also since you mentioned the reasons, this avoids the filesystem scan from `ListApps` and also allows `app ps` to include apps in unknown locations, with missing directories, or with broken metadata so it emphasises on robustness of command. :>

cmd/arduino-app-cli/app/ps.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// filterAppsByStatus keeps only apps that have been initialized at least once
+// (i.e. excludes StatusUninitialized), showing stopped apps only if showAll is set.
+func filterAppsByStatus(apps []orchestrator.AppInfo, showAll bool) []orchestrator.AppInfo {
+	res := make([]orchestrator.AppInfo, 0, len(apps))
+	for _, a := range apps {
+		if psVisibleStatuses[a.Status] || (showAll && a.Status == orchestrator.StatusStopped) {
+			res = append(res, a)
+		}
```
lucarin91: I would suggest using a switch instead; I found it more readable
```suggestion
      switch a.Status {
              case orchestrator.StatusStarting, orchestrator.StatusRunning, orchestrator.StatusStopping, orchestrator.StatusFailed:
                      res = append(res, a)
              case orchestrator.StatusStopped:
                      if showAll {
                              res = append(res, a)
                      }
              }
```

cmd/arduino-app-cli/app/ps.go
```diff
@@ -0,0 +1,97 @@
...
+		ctx,
+		servicelocator.GetDockerClient(),
+		servicelocator.GetAppIDProvider(),
+	)
+	if err != nil {
+		feedback.Fatal(err.Error(), feedback.ErrGeneric)
+	}
+
+	feedback.PrintResult(appPsResult{
+		Apps: filterAppsByStatus(apps, showAll),
```
lucarin91: You can directly f.Filter here instead define a new function
```go
		Apps: f.Filter(apps, func(a orchestrator.AppInfo) bool {...}
```

cmd/arduino-app-cli/app/ps_test.go
```diff
@@ -0,0 +1,51 @@
...
+	})
+}
+
+func namesOf(apps []orchestrator.AppInfo) []string {
+	names := make([]string, 0, len(apps))
+	for _, a := range apps {
+		names = append(names, a.Name)
+	}
+	return names
+}
```
lucarin91: I don't find this test very useful I think we can just remove it

internal/orchestrator/orchestrator.go
```diff
@@ -542,6 +543,60 @@ func ListApps(
...
+		slog.Warn("unable to load app metadata", slog.String("path", s.AppPath.String()), slog.String("error", err.Error()))
+		info.Name = s.AppPath.Base()
+		return info
+	}
+
+	info.Name = userApp.Name
+	info.Description = userApp.Descriptor.Description
+	info.Icon = userApp.Descriptor.Icon
+	return info
+}
```
lucarin91: This could be also semplify with `f.Map`
```suggestion
	return f.Map(appsStatus, func(s AppStatusInfo) AppInfo {
		info := AppInfo{Status: s.Status}

	if id, err := idProvider.IDFromPath(s.AppPath); err != nil {
		slog.Warn("unable to get app id", slog.String("path", s.AppPath.String()), slog.String("error", err.Error()))
	} else {
		info.ID = id
		info.Example = id.IsExample()
	}

	userApp, err := app.Load(s.AppPath)
	if err != nil {
		slog.Warn("unable to load app metadata", slog.String("path", s.AppPath.String()), slog.String("error", err.Error()))
		info.Name = s.AppPath.Base()
		return info
	}

	info.Name = userApp.Name
	info.Description = userApp.Descriptor.Description
	info.Icon = userApp.Descriptor.Icon
	return info
}), nil
```

cmd/arduino-app-cli/app/ps.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+type appPsResult struct {
+	Apps []orchestrator.AppInfo `json:"apps"`
+}
+
+func (r appPsResult) String() string {
+	t := table.NewWriter()
+	t.SetStyle(tablestyle.CustomCleanStyle)
+	t.AppendHeader(table.Row{"ID", "NAME", "STATUS"})
```
lucarin91: Maybe, for readability, it is better to move status right after the ID instead of the name
```suggestion
	t.AppendHeader(table.Row{"ID", "STATUS", "NAME"})
```

lucarin91 (commented):
Thanks for the updates, I added some suggestion to improve the code a little bit, but the implementation now is ok.

kirbx01:
> Thanks for the updates, I added some suggestion to improve the code a little bit, but the implementation now is ok.

Please check the final commit, I've applied all your suggestions and also did and removed the `ps_test.go` file.

cmd/arduino-app-cli/app/ps.go
```diff
@@ -0,0 +1,89 @@
...
+func (r appPsResult) String() string {
+	t := table.NewWriter()
+	t.SetStyle(tablestyle.CustomCleanStyle)
+	t.AppendHeader(table.Row{"ID", "STATUS", "NAME"})
+
+	for _, app := range r.Apps {
+		t.AppendRow(table.Row{
+			cmdutil.IDToAlias(app.ID),
+			app.Name,
+			app.Status,
```
lucarin91: You forgot to swap these two
```suggestion
            app.Status,
			app.Name,
```

lucarin91 (approved):

PR #613, opened 2026-09-01
Refactor(cli): make 'app list' a pure catalog view

### LINKED ISSUE #586: Refactor app list to make more usable (app ps)

kirbx01:
### Motivation
`app list` currently mixes two different concerns: browsing the app catalog (examples and user apps) and showing live status (running, stopped, failed, etc). As the catalog grew with `core-and-foundational` and `bricks` examples, the command became cluttered and hard to use, as reported in #586.

As outlined in the issue, the fix is to split the app catalog from the app status command: `app ps` (added in #612) for live status, and `app list` becoming a pure catalog view. **This PR implements the second half.** 

### Change description
I changed `app list` so it's a pure catalog browser, following the [layout specified](https://github.com/arduino/arduino-app-cli/issues/586#issuecomment-5436628770):

- By default, `app list` shows user apps only.
- `--examples` shows example apps only.
- `--all` / `-a` shows both user apps and example apps.

I removed the `ICON`, `STATUS`, and `EXAMPLE` columns from the table output and replaced them with a `DESCRIPTION` column. Status now lives in `app ps`, and the icon column was flagged in the issue as causing indentation issues across terminals.
I didn't need to change the orchestrator layer for this. `ListAppRequest` already exposes `ShowApps` and `ShowExamples`, so I only had to change how the CLI sets these flags based on the new command line flags, and update the table rendering in `list.go`.
This PR depends only on `main`, not on #612 (`app ps`), so either can be merged first without conflict.

### Demo
image

_The demo setup was updated to include an example app so the intended user/examples/all catalog behavior is visible._

### Additional Notes

Refs #586
This is the second half of the split proposed in #586, following #612 (`app ps`).

### Reviewer checklist

### DIFF
--- a/cmd/arduino-app-cli/app/list.go
+++ b/cmd/arduino-app-cli/app/list.go
@@ -21,26 +21,36 @@ import (
 )
 
 func newListCmd(cfg config.Configuration) *cobra.Command {
+	var showExamples bool
+	var showAll bool
 	var showBrokenApps bool
 
 	cmd := &cobra.Command{
 		Use:   "list",
-		Short: "List the Arduino apps",
+		Short: "List the Arduino apps catalog",
+		Long: "List the Arduino apps catalog.\n" +
+			"By default only user apps are shown. Use --examples to list the example apps, " +
+			"or --all to list both.\n" +
+			"To see the live status of the apps on the board, use 'app ps' instead.",
 		Run: func(cmd *cobra.Command, args []string) {
-			listHandler(cmd.Context(), cfg, showBrokenApps)
+			listHandler(cmd.Context(), cfg, showExamples, showAll, showBrokenApps)
 		},
 	}
 
+	cmd.Flags().BoolVar(&showExamples, "examples", false, "Only list the example apps")
+	cmd.Flags().BoolVarP(&showAll, "all", "a", false, "List both user apps and example apps")
 	cmd.Flags().BoolVarP(&showBrokenApps, "show-broken-apps", "", false, "Output a list of broken apps")
 	return cmd
 }
 
-func listHandler(ctx context.Context, cfg config.Configuration, showBrokenApps bool) {
+func listHandler(ctx context.Context, cfg config.Configuration, showExamples, showAll, showBrokenApps bool) {
 	res, err := orchestrator.ListApps(ctx,
 		servicelocator.GetDockerClient(),
 		orchestrator.ListAppRequest{
-			ShowExamples:                   true,
-			ShowApps:                       true,
+			// By default we only show user apps. --examples restricts the view to examples only,
+			// and --all shows both.
+			ShowApps:                       showAll || !showExamples,
+			ShowExamples:                   showAll || showExamples,
 			IncludeNonStandardLocationApps: true,
 		},
 		servicelocator.GetAppIDProvider(),
@@ -68,15 +78,13 @@ type appListResult struct {
 func (r appListResult) String() string {
 	t := table.NewWriter()
 	t.SetStyle(tablestyle.CustomCleanStyle)
-	t.AppendHeader(table.Row{"ID", "NAME", "ICON", "STATUS", "EXAMPLE"})
+	t.AppendHeader(table.Row{"ID", "NAME", "DESCRIPTION"})
 
 	for _, app := range r.Apps {
 		t.AppendRow(table.Row{
 			cmdutil.IDToAlias(app.ID),
 			app.Name,
-			app.Icon,
-			app.Status,
-			app.Example,
+			app.Description,
 		})
 	}
 	if r.showBrokenApps && len(r.BrokenApps) > 0 {


cmd/arduino-app-cli/app/list.go
```diff
@@ -68,15 +78,13 @@ type appListResult struct {
 func (r appListResult) String() string {
 	t := table.NewWriter()
 	t.SetStyle(tablestyle.CustomCleanStyle)
-	t.AppendHeader(table.Row{"ID", "NAME", "ICON", "STATUS", "EXAMPLE"})
+	t.AppendHeader(table.Row{"ID", "NAME", "DESCRIPTION"})
```
lucarin91, line 81: I would add a sign or something on the app that is set as the default one. Maybe just a `*` in the name (or ID).

cmd/arduino-app-cli/app/list.go
```diff
@@ -68,15 +78,13 @@ type appListResult struct {
...
 
 	for _, app := range r.Apps {
 		t.AppendRow(table.Row{
 			cmdutil.IDToAlias(app.ID),
 			app.Name,
-			app.Icon,
```
lucarin91, line 77: Maybe we can add back the icon and try to find a better way to ensure the rest of the table is alligned

lucarin91 (commented):
LGTM, I added some improvements. Anyway we should wait for the other PR before merging this otherwise we will lose the ability to get app status.

cmd/arduino-app-cli/app/list.go
kirbx01: okiee, please review PR #612  the requested changes are implemented and done and since these are minor tweaks i can subsequently proceed immediately after!

kirbx01:
Uh oh 😭, totally got confused i was implementing this one locally so meanwhile checking posted the comment meant for PR #612 . Really sorry for the inconvenience.

cmd/arduino-app-cli/app/list.go
kirbx01: Marked the default app with a '*' suffix on its name for this one

cmd/arduino-app-cli/app/list.go
kirbx01: The ICON column is restored and for alignment, it is handled by the go-pretty table library's automatic column-width sizing (measured via runewidth), so the rest of the table stays aligned regardless of icon width. Kept the change minimal: just added app.Icon back to the row and header. Refs commit 5781095d.

Xayton:
This is potentially a breaking change, but I would add, for the next few versions, a warning when running `app list`, saying:
```
Warning: the 'status' field has been removed from this list. You can use instead the new command 'app ps' to show the list of active apps.
```

This should also help people using AI agents, that are not aware yet of this change.
We should also update the [AGENTS.md](https://github.com/arduino/arduino-app-cli/blob/main/debian/arduino-app-cli/etc/arduino-app-cli/AGENTS.md) file.

kirbx01:
Hey @Xayton could you please check the latest commit, I've added the warning for the change that could potetntially break also that i have updated `AGENTS.md`, please revert if the AGENTS.md content is alright or i need to be more elaborate also that since the #612 PR has added 
```bash
arduino-app-cli app ps                                # starting/running/stopping/failed apps
``` 
Hence i have updated this too.

debian/arduino-app-cli/etc/arduino-app-cli/AGENTS.md
```diff
@@ -72,7 +72,9 @@ Command groups include `app`, `brick`, `model`, `monitor`, `properties`, `system
 **ID** like `user:my-app` or `examples:blink`):
 
 ```bash
-arduino-app-cli app list                              # apps + STATUS + IDs
+arduino-app-cli app list                              # app catalog + IDs
+arduino-app-cli app ps                                # starting/running/stopping/failed apps
```
Xayton:
```suggestion
arduino-app-cli app list                               # apps catalog + IDs
arduino-app-cli app list  --examples         # examples catalog + IDs
arduino-app-cli app ps                                # starting/running/stopping/failed apps
```

Xayton:
Hello @kirbx01, yes the latest changes are good, thanks.
While reviewing the feature I noticed a few more things. It would be nice if you could address those, and then we can merge, so that these changes will be part of the next release.

- rebase main, so that while testing we also have the `app ps` that has already been merged in main
- the description for some examples is too long: I would remove it from app list (I know we asked to add it earlier)
- app list (without additional flags) shows stopped examples. I think this is now a bug.
- Extend the warning message, mentioning also the other breaking change: the fact that app list will not show examples by default anymore. And to get those you need to pass the `--examples` flag. If we really want, we could also color the warning line yellow/orange.

kirbx01:
> Hello @kirbx01, yes the latest changes are good, thanks. While reviewing the feature I noticed a few more things. It would be nice if you could address those, and then we can merge, so that these changes will be part of the next release.
> 
> * rebase main, so that while testing we also have the `app ps` that has already been merged in main
> * the description for some examples is too long: I would remove it from app list (I know we asked to add it earlier)
> * app list (without additional flags) shows stopped examples. I think this is now a bug.
> * Extend the warning message, mentioning also the other breaking change: the fact that app list will not show examples by default anymore. And to get those you need to pass the `--examples` flag. If we really want, we could also color the warning line yellow/orange.

Please take another look i've resolved all the suggestions and implementation that had to be done, about colouring the warning, so to keep the CLI output formatting consistent and avoid introducing additional terminal-specific styling issues. The functional warning text is already clear without relying on color so it fits best of interest of issue to. :3

Xayton:
I have added a few more changes, for issues or improvements found during the review.
I have explained the changes in the 3 commits.

kirbx01:
@Xayton seeing all the commits i feel like this pr is done for now? I think we've resolved the aforementioned issues, are there any amendments to do?

lucarin91 (approved):

PR #621, opened 2026-09-09
fix: use project name instead of compose file path

lucarin91:
### Motivation

The app CLI cannot operate on apps that are missing the generated Compose files in the .cache directory. 

### Change description

By using the project name, we can always get logs and stop the application, also after a full removal of the .cache folder.

### Reviewer checklist


lucarin91:
Included in https://github.com/arduino/arduino-app-cli/pull/632.

PR #624, opened 2026-09-10
feat: added sketch artifact generation on app release build

cmaglie:
### Motivation

The app release lacks support for the pre-compiled sketch. This adds it.

### Change description

The release build now compiles the app sketch and stores the resulting firmware artifact in the archive. Apps without a sketch are unaffected.

To do this without duplicating the flashing logic, `compileUploadSketch` is split into three reusable parts: `initializeArduinoCli`, which sets up the arduino-cli server, loads the sketch and initialises the profile, `compileSketch`, and `uploadSketch`. The release build reuses the first two and asks arduino-cli to write the firmware to a file instead of uploading it to a board. The helpers in `sketch_flash.go` now take the FQBN they must use rather than deriving it from the platform, as the caller is the one that knows which board options apply.

### Additional Notes

Depends on arduino/arduino-cli#3287, which adds the "upload to file" capability. `go.mod` currently points at that.


internal/orchestrator/release.go
```diff
@@ -201,6 +205,11 @@ func BuildRelease(
 		return BuildReleaseResult{}, err
 	}
 
+	cb(StreamMessage{data: "building sketch", progress: &Progress{Name: "sketch", Progress: 20.0}})
+	if err := buildSketch(ctx, appToBuild, plat, prebuildDir, verbose, cb); err != nil {
```
Xayton: This line should use the stagedApp, not the original app passed by the user.


```suggestion
	if err := buildSketch(ctx, stagedApp, plat, prebuildDir, verbose, cb); err != nil {
```

internal/orchestrator/release.go
```diff
@@ -456,6 +465,57 @@ func buildPythonEnv(ctx context.Context, docker command.Cli, pythonImage string,
 	return nil
 }
 
+func buildSketch(ctx context.Context, appToBuild app.ArduinoApp, platform platform.Platform, destPath *paths.Path, verbose bool, cb func(StreamMessage)) error {
+	output := NewCallbackWriter(func(line string) {
+		cb(StreamMessage{data: line})
+	})
+
+	sketchPath, ok := appToBuild.GetSketchPath()
```
Xayton, line 475: An app without the sketch now fails to "release build". If the app has no sketch, the build step should be skipped.

internal/orchestrator/release.go
```diff
@@ -201,6 +205,11 @@ func BuildRelease(
 		return BuildReleaseResult{}, err
 	}
 
+	cb(StreamMessage{data: "building sketch", progress: &Progress{Name: "sketch", Progress: 20.0}})
```
Xayton:
```suggestion
	cb(StreamMessage{data: "building sketch", progress: &Progress{Name: "sketch", Progress: 80.0}})
```

internal/orchestrator/release.go
Xayton, line 102: Tests should also now test the case of an app with/without a sketch.

internal/orchestrator/release.go
```diff
@@ -456,6 +465,57 @@ func buildPythonEnv(ctx context.Context, docker command.Cli, pythonImage string,
...
+
+	srv, inst, err := initializeArduinoCli(ctx, sketchPath, output)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_, _ = srv.Destroy(ctx, &rpc.DestroyRequest{Instance: inst})
+	}()
+
+	fqbn := platform.FQBN + ":" + WaitForApp.String()
```
Xayton: Instead of appending `wait_linux_boot=app`, we should check if the platform supports it. Like it's done in `compileUploadSketch`.

internal/orchestrator/orchestrator.go
```diff
@@ -1058,28 +1108,23 @@ func compileUploadSketch(
...
+	w io.Writer,
+) error {
 	// build the sketch
-	buildPath := arduinoApp.SketchBuildPath()
-	if buildPath.NotExist() {
-		if err := buildPath.MkdirAll(); err != nil {
-			return fmt.Errorf("failed to create build directory: %w", err)
-		}
-	}
+	slog.Debug("compile and upload sketch", slog.String("fqbn", fqbn))
```
Xayton: Fix the log: it now only compiles the sketch.

internal/orchestrator/orchestrator.go
```diff
@@ -995,34 +995,84 @@ func compileUploadSketch(
...
+
+	srv, inst, err := initializeArduinoCli(ctx, sketchPath, w)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_, _ = srv.Destroy(ctx, &rpc.DestroyRequest{Instance: inst})
+	}()
+
+	menuOptions, err := GetPlatformMenuOptions(ctx, platform)
```
Xayton, line 959: Here the arduino-cli is instantiated twice: in the function and then inside GetPlatformMenuOptions.
Would be nice to avoid it.

internal/orchestrator/release.go
```diff
@@ -198,6 +202,15 @@ func BuildRelease(
 		return BuildReleaseResult{}, err
 	}
 
+	// The sketch is optional, as it is for the release manifest: an app made of
+	// python only ships without a firmware.
+	if _, hasSketch := appToBuild.GetSketchPath(); hasSketch {
+		cb(StreamMessage{data: "building sketch", progress: &Progress{Name: "sketch", Progress: 20.0}})
+		if err := buildSketch(ctx, appToBuild, plat, prebuildDir, verbose, cb); err != nil {
```
lucarin91: We should compile the app using the sketch source code in the staged directory to make sure we don't ship a compiled file different from the code we include

internal/orchestrator/release.go
Xayton: Actually its desired: we build using the cache of the existing app, then store the results in the destination folder (which is inside the stagedApp).

Xayton (approved):

PR #632, opened 2026-09-10
feat: use docker compose library

lucarin91:
### Motivation

Use the Docker Compose SDK instead of relying on the Docker Compose plugin installed on the board. This ensures we only depend on a Docker engine greater than 25.0.

### Change description

1. Move to the new Docker Compose SDK v5
2. Remove all docker compose subprocess calls
3. Add a new command `system check` that returns whether the system meets app-cli requirements
4. Use the project name when possible instead of the compose file (see https://github.com/arduino/arduino-app-cli/pull/621)

### Reviewer checklist


Xayton (approved):
Verified the overall code logic and tested with some apps.

PR #633, opened 2026-09-10
refactor: move docker stuff in dockerhelper

lucarin91:
### Motivation

Move all Docker utilities into the dockerhelper module.

### Reviewer checklist

### DIFF
--- a/internal/dockerhelper/images.go
+++ b/internal/dockerhelper/images.go
@@ -12,7 +12,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"log/slog"
+	"os"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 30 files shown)


internal/dockerhelper/images.go
```diff
@@ -0,0 +1,239 @@
...
+
+// ImagePrefixes states which images are ours, past ones included: what to pull, and
+// what a cleanup may remove.
+var ImagePrefixes = []string{
+	"ghcr.io/bcmi-labs/",
+	"public.ecr.aws/arduino/",
+	"ghcr.io/arduino/",
+	"influxdb",
+	"artifacts.codelinaro.org/iot-solutions-microservices/",
+}
```
Xayton: This ImagePrefixes is mutable, and two tests (TestPullImages , TestExtractImagesFromCompose) modify it while testing. Could cause race condition issues.

internal/dockerhelper/images.go
```diff
@@ -0,0 +1,239 @@
...
+		allLayers = append(allLayers, layers...)
+	}
+	// Shared layers are counted once, so this is the real download size.
+	totalBytes := sumUniqueLayers(allLayers)
+	slog.Info("total docker images download size", "bytes", totalBytes)
+
+	// The engine may keep its images on another host, and then the space of this one
+	// says nothing: what cannot be checked does not stop a download.
+	if freeSpace, err := dockerFreeSpace(); err != nil {
+		slog.Warn("cannot read the free space of the docker partition", "error", err)
```
Xayton: Now that PullImages is used in all cases, this has been changed from an error to a warning.

This leaves open a narrow case: on a board (where docker is local) the docker folder has been configured to another location; if that folder is actually full, the dockerFreeSpace will return an error (because the default `/var/lib/docker` folder is missing or unaccessible), a warn will be logged and the process will continue. Breaking badly because the disk is full.

One possible solution: query the location of the docker folder (`info, err := docker.Info(ctx) // info.DockerRootDir`). If it's local and different from the default, check that and restore the error. If the docker engine is remote skip the check (and warn in this case).

Xayton (approved):

mirkoCrobu (approved):

PR #635, opened 2026-09-11
Fastrpc devices support - main container support

rjtokenring:
### Motivation

Add minimum requirement to have fastrpc-* and dma devices accessible from main python container

### Reviewer checklist


internal/orchestrator/compose_template.go
```diff
@@ -173,11 +173,20 @@ func generateComposeTemplate(
...
+
 	// Mounted only where the board has them.
 	optionalMounts := slices.Concat(
 		[]string{"/run/udev:ro", "/run/user/1000/pipewire-0"},
 		// camx CSI cameras are accessed through the cam_server socket and a host userspace library
 		[]string{"/run/cam_server", "/usr/lib/libcamera_metadata.so.0.1.0"},
+		// DSP binaries, at a stable path in the container wherever the host keeps them
+		[]string{dspPath + ":/run/host-qcom:ro"},
+		// the board model, where it is
+		[]string{"/sys/firmware/devicetree/base/model:ro"},
```
lucarin91, line 182: Why do you need the base model name inside the container?

internal/orchestrator/compose_template.go
```diff
@@ -173,11 +173,20 @@ func generateComposeTemplate(
 		},
 	}
 
+	// The DSP binaries live in /usr/share/qcom, or wherever the host says. Resolved
+	// here and not as a ${VAR} of the template: a mount is only written where its
+	// path exists, which is checked before the template is interpolated.
+	dspPath := cmp.Or(os.Getenv("HOST_DSP_INSTALLATION_PATH"), "/usr/share/qcom")
```
lucarin91: We should check the DSP path on the host, not in the builder.
If you know some locations of that, you can directly add them to the `optionalMounts` below, which will automatically mount the ones existing on the running host.

internal/orchestrator/compose_template.go
```diff
@@ -225,7 +234,7 @@ func generateComposeTemplate(
 		}
 	}
 
-	deviceDrivers := []string{"drm", "dma_heap", "media", "video4linux", "alsa", "ttyUSB", "ttyACM"}
+	deviceDrivers := []string{"drm", "dma_heap", "media", "video4linux", "alsa", "ttyUSB", "ttyACM", "misc"}
```
lucarin91: `misc` major will include a lot of different devices, not only the fastrpc one. Maybe we should restrict this to only the FastRPC ones. If it is ok to you, I can push a patch for that.

lucarin91 (changes requested):

internal/orchestrator/compose_template.go
rjtokenring: it's required by fastrpc, mandatory

lucarin91 (approved):

PR #643, opened 2026-09-15
fix: print start/restart error messages verbatim

### LINKED ISSUE #77: Do not use the the Caser for language-specific casing in the cli output
## Motivation
The arduino-app-cli uses the `"golang.org/x/text/cases"` to print the error to the console with capitalize words.

This causes an issue when a word MUST not change.

For example, in the following case, the `ARDUINO_DEVICE_ID` is a configuration variable that must not be changed in the error exposed to the user.

Example:
Input: `variable \"ARDUINO_DEVICE_ID\" is required by brick \"arduino:arduino_cloud\"`
Output: `[ERROR] Variable "Arduino_device_id" Is Required By Brick "Arduino:arduino_cloud"`

lurenm111:
## Summary

The `start` and `restart` commands previously passed error messages through `cases.Title(language.AmericanEnglish)`, which capitalized every word and corrupted identifiers such as the `ARDUINO_DEVICE_ID` environment variable (reported in #77).

Before:
```
[ERROR] Variable "Arduino_device_id" Is Required By Brick "Arduino:arduino_cloud"
```

After:
```
[ERROR] variable "ARDUINO_DEVICE_ID" is required by brick "arduino:arduino_cloud"
```

## Changes

- `cmd/arduino-app-cli/app/start.go`: remove `cases.Title` wrapping, print error verbatim
- `cmd/arduino-app-cli/app/restart.go`: same fix

This matches the convention already used by the `build` command in the same directory.

## Verification

- Behavior reproduction confirms `ARDUINO_DEVICE_ID` is preserved after the fix
- `go build` (linux/amd64) passes for both patched and pristine baseline
- `go vet ./cmd/arduino-app-cli/app/` passes
- `gofmt -l` reports no changes
- `go mod tidy` leaves go.mod/go.sum unchanged (`x/text` still used by `cmd/gendoc`)

Closes #77


lucarin91:
@lurenm111 can you sign the cla?

Xayton (approved):

lucarin91:
We cannot merge the PR without the CLA signed, so we will close this PR.

PR #645, opened 2026-09-16
feat(api): add App Releases build API

MatteoPologruto:
### Motivation

This PR adds the API to build an app-release. The `/v1/apps/${APP_ID}/build` is a `POST` that builds the release and returns it as an `application/gzip`. The SSE events of each build related to an app are available at the endpoint `/v1/apps/${APP_ID}/build/events`, eventually tagged with the corresponding `build_id` to be later filtered.

The API can be tested by the following request:

```
export APP_ID="ZXhhbXBsZXM6aW5zcGlyYXRpb25hbC9ibGluaw" && \
curl -N -X POST "http://127.0.0.1:8800/v1/apps/${APP_ID}/build" \
 -H "Content-Type: application/json" \
 -d '{
  "target": "ventunoq",
  "include_data": true,
  "build_id": "1234",
  "release_label": "1.0"
  "release_notes": "🚀 This is a test note\n This is a simple test"
 }'

```
Tested with wring/missing options.

### Reviewer checklist


internal/api/api.go
```diff
@@ -76,6 +76,8 @@ func NewHTTPRouter(
 	mux.Handle("GET /v1/apps/{appID}/logs", handlers.HandleAppLogs(dockerClient, idProvider, bricksIndex, cfg))
 	mux.Handle("POST /v1/apps/{appID}/start", handlers.HandleAppStart(dockerClient, provisioner, modelsIndex, bricksIndex, servicesIndex, idProvider, cfg, platform))
 	mux.Handle("POST /v1/apps/{appID}/stop", handlers.HandleAppStop(dockerClient, idProvider, platform, cfg))
+	mux.Handle("POST /v1/apps/{appID}/build", handlers.HandleAppBuild(dockerClient, provisioner, idProvider, cfg))
+	mux.Handle("GET /v1/apps/{appID}/build/{artifactID}", handlers.HandleAppBuildArtifact(idProvider))
```
lucarin91, line 80: I definitely prefer to have the download inline with the build; otherwise, we need to handle cleanup of the artifacts.

lucarin91 (dismissed):

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+			broker.Publish(appKey, render.SSEEvent{Type: "error", Data: buildErrorEvent{BuildID: buildID, Code: code, Message: err.Error()}})
+			render.EncodeResponse(w, status, models.ErrorResponse{Details: err.Error()})
+			return
+		}
+
+		archivePath := paths.New(result.Archive)
+		defer func() { _ = archivePath.Remove() }()
+
+		broker.Publish(appKey, render.SSEEvent{Type: "done", Data: buildDoneEvent{BuildID: buildID, Name: result.Name, Target: result.Target}})
```
mirkoCrobu: nit
publish done after the Exist() check; otherwise, the stream can report success while the response is a 500.

internal/releasebuild/event.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"sync"
+
+	"github.com/arduino/arduino-app-cli/internal/render"
+)
+
+// EventBroker fans out build progress events to per-app subscribers, so a
+// single app-wide SSE stream can observe every build of that app. Each event
+// carries the build id it belongs to, letting a subscriber filter by build.
+type EventBroker struct {
```
mirkoCrobu, line 17: do you think we could add some tests for the brocker?

internal/api/api.go
```diff
@@ -76,6 +78,8 @@ func NewHTTPRouter(
 	mux.Handle("GET /v1/apps/{appID}/logs", handlers.HandleAppLogs(dockerClient, idProvider, bricksIndex, servicesIndex, cfg))
 	mux.Handle("POST /v1/apps/{appID}/start", handlers.HandleAppStart(dockerClient, provisioner, modelsIndex, bricksIndex, servicesIndex, idProvider, cfg, platform))
 	mux.Handle("POST /v1/apps/{appID}/stop", handlers.HandleAppStop(dockerClient, idProvider, platform, cfg))
+	mux.Handle("POST /v1/apps/{appID}/build", handlers.HandleAppBuild(dockerClient, provisioner, idProvider, cfg, buildEvents))
```
mirkoCrobu, line 81: could we add test e2e for this one?

MatteoPologruto: I added some tests

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/arduino/arduino-app-cli/internal/orchestrator/appid"
+	"github.com/arduino/arduino-app-cli/internal/orchestrator/config"
+	"github.com/arduino/arduino-app-cli/internal/platform"
+	"github.com/arduino/arduino-app-cli/internal/releasebuild"
+	"github.com/arduino/arduino-app-cli/internal/render"
+)
+
+// buildRequest is the JSON body of a build request.
+type buildRequest struct {
+	BuildID      string `json:"build_id" description:"optional build identifier used to tag the build events stream"`
```
mirkoCrobu, line 33: should it be mandatory?

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		if len(body) > 0 {
+			if err := json.Unmarshal(body, &buildReq); err != nil {
+				slog.Error("unable to decode app build request", slog.String("error", err.Error()))
+				render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "unable to decode app build request"})
+				return
+			}
+		}
+
+		buildID := buildReq.BuildID
```
mirkoCrobu, line 104: What happens if it is empty?
Currently, it is optional, so should we make it mandatory, or maybe we could generate an id by ourself?

mirkoCrobu: In this case sse event doesn't have any identifier

internal/api/handlers/app_build.go
MatteoPologruto: It should not, the whole point is having the caller generate it if there is a reason to track (and filter) the SSE events for a specific app release build, while limiting the SSE connections. The caller should be free to omit it.

internal/api/handlers/app_build.go
MatteoPologruto: If it is empty, it's not possible to track a specific build, but the untagged events are still sent to the `/events` endpoint. Even if we generated it internally, the caller would be unaware of it, so it would not help in tracking specific events. I think it should be the caller's responsibility to use a `build_id` and store it somewhere if it is interested in tracking the SSE events (also see the other comment).

internal/releasebuild/event.go
MatteoPologruto: Added unit tests for the broker

mirkoCrobu (approved):

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,203 @@
...
+		BuildID string `json:"build_id,omitempty"`
+		Name    string `json:"name"`
+		Target  string `json:"target"`
+	}
+	buildErrorEvent struct {
+		BuildID string            `json:"build_id,omitempty"`
+		Code    render.SSEErrCode `json:"code"`
+		Message string            `json:"message,omitempty"`
+	}
+)
```
lucarin91, line 62: can you move this close to the event handler?

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+
+		appToBuild, err := app.Load(id.ToPath())
+		if err != nil {
+			slog.Error("unable to load the app", slog.String("error", err.Error()), slog.String("path", id.String()))
+			render.EncodeResponse(w, http.StatusInternalServerError, models.ErrorResponse{Details: "unable to load the app"})
+			return
+		}
+
+		defer r.Body.Close()
```
lucarin91, line 87: Handlers doesn't need to close the body the server will do it
```suggestion
```

>	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,203 @@
...
+			render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "unable to read app build request"})
+			return
+		}
+		if len(body) > 0 {
+			if err := json.Unmarshal(body, &buildReq); err != nil {
+				slog.Error("unable to decode app build request", slog.String("error", err.Error()))
+				render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "unable to decode app build request"})
+				return
+			}
+		}
```
lucarin91, line 102: we we didn't directly unmarchall?
```suggestion
			if err := json.Unmarshal(r.Body, &buildReq); err != nil {
				slog.Error("unable to decode app build request", slog.String("error", err.Error()))
				render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: "unable to decode app build request"})
				return
			}
```

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,203 @@
...
+
+		buildID := buildReq.BuildID
+
+		if buildReq.Target != "" {
```
lucarin91, line 106: target sould be always define

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,203 @@
...
+			if _, ok := platform.ForBoard(buildReq.Target); !ok {
+				render.EncodeResponse(w, http.StatusBadRequest, models.ErrorResponse{Details: fmt.Sprintf("the field 'target' must be one of %s", strings.Join(platform.SupportedBoards(), ", "))})
+				return
+			}
+		}
+
+		// Finished archives live here
+		// TODO 2 to be defined how it should work. handle deletion/TTL and so on...
+		artifactsDir := buildArtifactsDir()
```
lucarin91, line 115: I would remove the comment and remove the function.
```suggestion
		artifactsDir := paths.New(os.TempDir(), "build-artifacts")
```
In a future PR, we should also attempt to avoid using this file and stream the archive directly to the caller

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,203 @@
...
+
+		req := orchestrator.BuildReleaseRequest{
+			Target:       buildReq.Target,
+			ReleaseLabel: buildReq.ReleaseLabel,
+			Notes:        buildReq.ReleaseNotes,
+			IncludeData:  buildReq.IncludeData,
+			Output:       artifactsDir,
+			Overwrite:    true,
+		}
+
```
lucarin91, line 130: nit
```suggestion
```

internal/api/handlers/app_build.go
```diff
@@ -0,0 +1,203 @@
...
+		result, err := orchestrator.BuildRelease(r.Context(), dockerClient, provisioner, appToBuild, req, cfg, func(item orchestrator.StreamMessage) {
+			// A StreamMessage may carry a progress value, an info message, or both,
+			// so publish each independently to avoid dropping either one.
+			if p := item.GetProgress(); p != nil {
+				broker.Publish(render.SSEEvent{Type: "progress", Data: buildProgressEvent{BuildID: buildID, Name: p.Name, Progress: p.Progress}})
+			}
+			if item.GetData() != "" {
+				broker.Publish(render.SSEEvent{Type: "message", Data: buildMessageEvent{BuildID: buildID, Message: item.GetData()}})
+			}
```
lucarin91, line 139: I think we should either try to reuse the update event broker or make the builderBrocker do the conversion between streamMessage and sseevents

internal/api/handlers/app_build.go
MatteoPologruto: I agree, I wanted to keep this PR simpler

internal/api/handlers/app_build.go
MatteoPologruto: If this isn't defined, isn't it set to the connected board down the line?

internal/api/handlers/app_build.go
MatteoPologruto: These are actually used by the build handler to send the events, even if they are seen through the events endpoint

MatteoPologruto: I actually moved them inside the broker.

internal/api/handlers/app_build.go
MatteoPologruto: The release build broker is now handling the conversion from stream message to SSE event

internal/api/handlers/app_build.go
lucarin91: right

lucarin91 (approved):

cmd/gendoc/docs.go
```diff
@@ -567,6 +567,72 @@ Contains a JSON object with the details of an error.
...
+
+**Event 'message'**:
+Contains a JSON object with an informational message.
+'event: message'
+'data: {"build_id":"abc","message":"building the python environment..."}'
+
+**Event 'done'**:
+Contains a JSON object with the built release facts.
+'event: done'
+'data: {"build_id":"abc","name":"my-app","target":"unoq"}'
```
martacarbone: maybe adding the app selector is an example near the real usage. user:my_app

martacarbone (approved):

PR #647, opened 2026-09-17
fix: system update handle dpkg lock

### LINKED ISSUE #642: system update should handle dkpg lock
If another apt update is in progress we aren't returning a reasonable error:

```
E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 5139 (unattended-upgr)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
Upgrade completed. Restarting ...
Reconnecting to board...
Log stream error: {}
Log stream error: {}
Log stream error: {}
```

We should check if the apt is locked and return, also in the apt check operation, an error, asking to retry later.

martacarbone:
### Motivation

Apply some fixes for a system update.
- do not restart the system if not required (apt failures)
- returns a more meaningful error if the apt lock is held by another process
- test the apt lock when check for updates (only for debian updates)

To test one can hold the lock by:
```
 sudo python3 -c '
import fcntl, time, os, sys

# Set process name so apt sees "unattended-upgr"
try:
    import setproctitle
    setproctitle.setproctitle("unattended-upgr")
except ImportError:
    sys.argv[0] = "unattended-upgr"

f = open("/var/lib/dpkg/lock-frontend", "w")
fcntl.lockf(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
print(f"Locked /var/lib/dpkg/lock-frontend by PID {os.getpid()} (unattended-upgr)")
time.sleep(3600)
```
### Reviewer checklist


internal/update/apt/service.go
```diff
@@ -188,6 +188,20 @@ func runUpdateCommand(ctx context.Context) error {
 	return nil
 }
 
+// checkAptLockHeld probes whether the dpkg lock is held by another process by running
+// an apt-get install for a package that does not exist
+func checkAptLockHeld(ctx context.Context) error {
+	cmd, err := paths.NewProcess([]string{debianFrontend}, "sudo", "apt-get", "install", "--assume-no", "non-existent-package-probe")
```
lucarin91, line 196: ~Is there something without sudo? I would prefer to avoid adding other staff to sudoers.~ 
This doesn't need an extra sudoers entry, so I think it is ok

internal/update/apt/service.go
```diff
@@ -187,6 +190,25 @@ func runUpdateCommand(ctx context.Context) error {
...
+// by running an apt-get install for a package that does not exist.
+func checkAptLockHeld(ctx context.Context) error {
+	cmd, err := paths.NewProcess([]string{debianFrontend}, "sudo", "apt-get", "install", "--assume-no", "non-existent-package-probe")
+	if err != nil {
+		return err
+	}
+	out, err := cmd.RunAndCaptureCombinedOutput(ctx)
+	if err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) &&
```
lucarin91, line 203:
```suggestion
		if exitErr, ok := errors.AsType[*exec.ExitError](err); ok &&
```

lucarin91 (approved):

internal/update/apt/service_test.go
```diff
@@ -15,6 +17,24 @@ import (
...
+		require.Equal(t, 100, exitErr.ExitCode())
+	})
+
+	t.Run("lock text is treated as lock held", func(t *testing.T) {
+		out := "E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)"
+		err := errors.New("boom")
+		require.True(t, strings.Contains(strings.ToLower(out), "lock"))
+		require.NotNil(t, err)
+	})
+}
```
lucarin91: this doesn't see a very useful test

lucarin91 (approved):

PR #652, opened 2026-09-18
fix(pkg/board): return file not found error

lucarin91:
### Motivation

`GetOSImageVersion` incorrectly reports that a board has RO image if there was a connection error. This is because ReadFile doesn't distinguish between a general error and a ErrNotExist error.

### Change description

Make `ReadFile` api return `ErrNotExist`

### Reviewer checklist


pkg/board/remote/parser.go
```diff
@@ -8,11 +8,53 @@ package remote
...
+		if err == nil {
+			return nil
+		}
+
+		msg := strings.TrimSpace(string(stderr))
+		switch {
+		case strings.Contains(msg, "No such file or directory"):
+			return fmt.Errorf("%w: %s", fs.ErrNotExist, msg)
+		case strings.Contains(msg, "Permission denied"):
+			return fmt.Errorf("%w: %s", fs.ErrPermission, msg)
```
Xayton: This depends on the locale on the board. Options:
- keep as is
- force locale to basic English (LC_ALL=C)
- also use the exit status of cat (exits with 1 in case of error)

Xayton (commented):
### Partial read errors
if the file is read partially and then there is an error (truncated read, I/O error, connection drop mid-transfer, remote killed) ParseReadOutput will not look at the exit() value and the error will only be visible via the Close() function.

Problem is GetOSImageVersion ignores any error returned by Close() (`defer f.Close()`).

- Check the exit() value also after the final read
- Do not ignore the Close() error in GetOSImageVersion

PR #662, opened 2026-09-21
fix(pkg/board): report failures of ExecAsRoot

### LINKED ISSUE #320: ExecAsRoot should return a specific error if the password is not correct
https://github.com/arduino/arduino-app-cli/pull/315

Xayton:
ExecAsRoot discarded stderr and ignored the closer, which carries the exit status of the command, so it always reported success. A wrong password, or any failing command, was silently treated as a successful run by EnableNetworkMode and DisableNetworkMode.

The exit status is now returned. A wrong password is not told apart from a failing command, because sudo exits with the same status in both cases.

### How to test

```
arduino-app-cli system network-mode status
arduino-app-cli system network-mode disable # put a WRONG password. it should return an error
echo $?
arduino-app-cli system network-mode status # status should not have changed

arduino-app-cli system network-mode disable # put the CORRECT password. it should work
```

Closes #320


lucarin91 (approved):
I prefer to return a specific authentication fail error, but at least this returns some errors.

Xayton:
> I prefer to return a specific authentication fail error, but at least this returns some errors.

Yes me too, but returning a specific error would involve parsing the text output of sudo, which is also affected by locale settings. So I thought this was already an improvement, without adding too much complexity.

